PR #21, opened 2021-07-21
Generate Apple Silicon wheels

oscargm98:


FrancescAlted:
This has stalled for too much time.  Closing (we are in [Closember](https://closember.org) after all).

PR #22, opened 2021-07-28
Add c-blosc2 library functionalities

martaiborra:
Add the capability of creating and working with a super-chunk.


FrancescAlted (approved):
Looks great to me!

PR #27, opened 2021-08-09
Add aarch64 wheel build support

odidev:
Added aarch64 wheel build support.
@FrancescAlted Could you please review this PR?

### DIFF
--- a/.github/workflows/cibuildwheels.yml
+++ b/.github/workflows/cibuildwheels.yml
@@ -21,12 +21,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        arch: [x86, amd64]
+        arch: [auto, aarch64]
         exclude:
-          - os: ubuntu-latest
-            arch: x86
+          - os: windows-latest
+            arch: aarch64
           - os: macos-latest
-            arch: x86
+            arch: aarch64
 
     steps:
       - name: Checkout repo
@@ -38,6 +38,10 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: '3.7'
+
+      - name: Set up QEMU
+        if: ${{ matrix.arch == 'aarch64' }}
+        uses: docker/setup-qemu-action@v1
 
       - name: Install Ninja
         uses: seanmiddleditch/gha-setup-ninja@master
@@ -54,7 +58,7 @@ jobs:
           python -m pip install -r requirements-tests.txt
 
       - name: Build wheels (Windows x86)
-        if: ${{ runner.os == 'Windows' && matrix.arch == 'x86'}}
+        if: ${{ runner.os == 'Windows' }}
         run: |
           python -m pip install --upgrade pip
           python -m pip install cibuildwheel
@@ -72,7 +76,7 @@ jobs:
           arch: amd64
 
       - name: Build wheels (Windows / amd64)
-        if: ${{ runner.os == 'Windows' && matrix.arch == 'amd64'}}
+        if: ${{ runner.os == 'Windows' }}
         run: |
           python -m pip install --upgrade pip
           python -m pip install cibuildwheel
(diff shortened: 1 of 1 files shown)


.github/workflows/cibuildwheels.yml
```diff
@@ -21,12 +21,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        arch: [x86, amd64]
+        arch: [auto, aarch64]
```
FrancescAlted, line 24: What `auto` means here?  Why `x86` and `amd64` are gone?

.github/workflows/cibuildwheels.yml
```diff
@@ -54,7 +58,7 @@ jobs:
           python -m pip install -r requirements-tests.txt
 
       - name: Build wheels (Windows x86)
-        if: ${{ runner.os == 'Windows' && matrix.arch == 'x86'}}
+        if: ${{ runner.os == 'Windows' }}
```
FrancescAlted, line 61: Why you are removing the `matrix.arch == 'x86'` condition?  Is that related with `auto`?

.github/workflows/cibuildwheels.yml
```diff
@@ -72,7 +76,7 @@ jobs:
           arch: amd64
 
       - name: Build wheels (Windows / amd64)
-        if: ${{ runner.os == 'Windows' && matrix.arch == 'amd64'}}
+        if: ${{ runner.os == 'Windows' }}
```
FrancescAlted, line 79: Ditto.

FrancescAlted (commented):
Sorry for the delay on this.  I got diverted and forgot about this.

I would appreciate if you can shed some light on my comments.  Thanks in advance.

.github/workflows/cibuildwheels.yml
hirnimeshrampuresoftware: Here `auto` takes all the available architecture for the given OS. Hence it will automatically take `x86` and `amd64`. Also `aarch64` is not available under `auto`, hence it is provided separately.

.github/workflows/cibuildwheels.yml
hirnimeshrampuresoftware: Since `x86` and `amd64` are replaced with `auto`, this condition is removed. Also, `CIBW_BUILD` is provided in the `env` which will build the required wheel.

.github/workflows/cibuildwheels.yml
odidev: Here `auto` takes all the available architecture for the given OS. Hence it will automatically take `x86` and `amd64`. Also `aarch64` is not available under `auto`, hence it is provided separately.

.github/workflows/cibuildwheels.yml
odidev: Since `x86` and `amd64` are replaced with `auto`, this condition is removed. Also, `CIBW_BUILD` is provided in the `env` which will build the required wheel.

.github/workflows/cibuildwheels.yml
FrancescAlted: Got it.

.github/workflows/cibuildwheels.yml
FrancescAlted: Understood.  So I'd say that the section below ('Build wheels (Windows / amd64)') is redundant.

FrancescAlted: If I understand this correctly, this also means that this should be producing wheels for `aarch64` arch too.

odidev: No, It produces only for the supported architectures of a build machine.

odidev: > Understood. So I'd say that the section below ('Build wheels (Windows / amd64)') is redundant.

Yes, this can be removed.

FrancescAlted: Thanks!

.github/workflows/cibuildwheels.yml
FrancescAlted: This section would be removed.

FrancescAlted:
Due to reasons explained in https://github.com/Blosc/python-blosc2/pull/47#issuecomment-1225633013, the aarch64 wheels will be disabled for now.

Issue #25, opened 2021-07-30
Be consistent with Caterva naming for codecs/filters


FrancescAlted:
Right now, python-blosc2 reuses the names for codecs/filters of the C-Blosc2 library.  However, it would make more sense to introduce a more pythonic way to do it, consistent with the ecosystem.  In particular, the convention followed by Caterva makes more sense:

https://github.com/Blosc/python-caterva/blob/master/caterva/utils.py#L14-L33

Let's do that.  Even if this represents a change in API, python-blosc2 is still in alpha, so this is not a big issue yet.

FrancescAlted:
Fixed by #31

PR #31, opened 2021-10-02
Filter and codec enums

FrancescAlted:
This PR introduces new `Codec` and `Filter` Enum classes.  This is handier for the users to check for available filters and codecs.


FrancescAlted:
Fixes #25.

PR #33, opened 2022-06-29
Add open function

martaiborra:


tests/test_open.py
```diff
@@ -0,0 +1,76 @@
+########################################################################
+#
+#       Created: May 26, 2021
```
FrancescAlted, line 3: This has been created later, right? ;-). At any rate, I don't think this field has much sense, and I'd remove it from all files.

blosc2/SChunk.py
```diff
@@ -274,7 +294,58 @@ def update_data(self, nchunk, data, copy):
...
             If some problem was detected.
         """
+        blosc2_ext._check_access_mode(self.urlpath, self.mode)
         return super(SChunk, self).update_data(nchunk, data, copy)
 
     def __dealloc__(self):
         super(SChunk, self).__dealloc__()
+
+
+def open(urlpath, mode="a",  **kwargs):
```
FrancescAlted, line 304: Two spaces before `**kwargs`

FrancescAlted (approved):
I just have some cosmetic comments.  Otherwise this looks great.

blosc2/SChunk.py
martaiborra: Fixed in 4b00b0a790da9cf1578b13a6cdf761552a28fb92.

PR #36, opened 2022-07-15
Close file and remove unnecessary 'else' clause

DimitriPapadopoulos:
File handles are a limited resource.

The '`for`' statement has a redundant '`else`' as no '`break`' is present in the body.

### DIFF
--- a/blosc2/core.py
+++ b/blosc2/core.py
@@ -603,8 +603,9 @@ def os_release_pretty_name():
                     return value
         except IOError:
             pass
-    else:
-        return None
+        finally:
+			f.close()
+    return None
 
 
 def print_versions():


blosc2/core.py
```diff
@@ -603,8 +603,9 @@ def os_release_pretty_name():
                     return value
         except IOError:
             pass
-    else:
-        return None
+        finally:
+			f.close()
```
FrancescAlted, line 607: Too much indentation

FrancescAlted:
Fixed in 48847b7

blosc2/core.py
DimitriPapadopoulos: Rather tabs instead of spaces, sorry about that.

PR #41, opened 2022-07-16
Use min bultin function

DimitriPapadopoulos:


blosc2/__init__.py
```diff
@@ -78,8 +78,7 @@ class Filter(Enum):
 # Internal Blosc threading
 nthreads = ncores = detect_number_of_cores()
 # Protection against too many cores
-if nthreads > 8:
-    nthreads = 8
+nthreads = min(nthreads, 8)
```
FrancescAlted, line 81: IMO this is not really necessary

FrancescAlted (commented):
Declining this.

blosc2/__init__.py
DimitriPapadopoulos: Well, I believe it is more readable, but i doesn't extend well to cases such as:
```python
if nthreads > 8:
    nthreads = 8
elif nthreads < 2:
    nthreads = 2
```

PR #47, opened 2022-08-20
Fix invalid pointer to urlpath argument

### LINKED ISSUE #46: Segfaults when run in Python's "development mode"
On Windows, `blosc2.compress` and `blosc2.compress2` reproducible crash when run on any current Python version in "development mode" with blosc2 0.3.1 official or local builds, e.g.:

```
> py -3.10-64 -X dev -c"from blosc2 import compress;compress(b'\x00')"
Debug memory block at address p=00000245606EFD60: API 'o'
    66 bytes originally requested
    The 7 pad bytes at p-7 are FORBIDDENBYTE, as expected.
    The 8 pad bytes at tail=00000245606EFDA2 are not all FORBIDDENBYTE (0xfd):
        at tail+0: 0x00 *** OUCH
        at tail+1: 0x00 *** OUCH
        at tail+2: 0xfd
        at tail+3: 0xfd
        at tail+4: 0xfd
        at tail+5: 0xfd
        at tail+6: 0xfd
        at tail+7: 0xfd
    Data at p: 00 00 00 00 00 00 00 00 ... 00 00 00 00 00 00 00 00

Enable tracemalloc to get the memory block allocation traceback

Fatal Python error: _PyMem_DebugRawFree: bad trailing pad byte
Python runtime state: initialized

Current thread 0x00003b64 (most recent call first):
  File "<string>", line 1 in <module>

Extension modules: blosc2.blosc2_ext (total: 1)
```

cgohlke:
Fixes segfaults during tests on Windows in "development mode".

The C pointer `storage.urlpath` was pointing to the value of a Python bytes instance, which got deleted while the pointer  was still in use.

I noticed that the return values of some C function calls are not checked for errors and c-blosc does not check some pointer arguments for NULL...

### DIFF
--- a/blosc2/blosc2_ext.pyx
+++ b/blosc2/blosc2_ext.pyx
@@ -22,8 +22,6 @@ from enum import Enum
 import blosc2
 
 
-cdef extern int blosc2_remove_urlpath(const char *path)
-
 cdef extern from "<stdint.h>":
     ctypedef   signed char  int8_t
     ctypedef   signed short int16_t
@@ -357,6 +355,7 @@ cdef extern from "blosc2.h":
     void blosc1_set_schunk(blosc2_schunk *schunk)
 
     int blosc2_remove_dir(const char *path)
+    int blosc2_remove_urlpath(const char *path)
 
 
 MAX_TYPESIZE = BLOSC_MAX_TYPESIZE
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Looks good!  Thanks!

cgohlke:
Does this fix the problems on aarch64 and musl platforms?

FrancescAlted:
Well, it is getting better.  Now, I am getting only an [Error: The operation was canceled](https://github.com/Blosc/python-blosc2/runs/7937966924?check_suite_focus=true#step:9:10821), so definitely your patch helped.

Now, I am guessing that the job for building wheels for arm64 is being cancelled because it is taking too much time.  Maybe splitting this job in two (say one for glibc and the other for musl), should help.

cgohlke:
Do numpy and psutil provide wheels for or even support musllinux_aarch64? Maybe those packages are failing to install or build until CI times out?

FrancescAlted:
> Do numpy and psutil provide wheels for or even support musllinux_aarch64? Maybe those packages are failing to install or build until CI times out?

That's a good point.  Apparently the numpy team is not building [musllinux_aarch64 yet](https://github.com/numpy/numpy/issues/20089), and is probably happening here is that CI is trying to build one from scratch.

Unfortunately, somehow the default is to build wheels with musllinux support (I suppose this is related with the incremental deployment of https://peps.python.org/pep-0656/).  This, together with the lack of a native aarch64 platform in GH Actions, makes this *very* slow, and hence, prone to be cancelled due to timeout.  So I think for now it is better to disable the wheels for aarch64 completely.  When NumPy would have native wheels for aarch64 and muslinux this can be tackled again.

PR #60, opened 2022-09-19
Cframe

martaiborra:
Add method to get frame from a SChunk
Add function to get a SChunk from a frame
Reformat doc titles from "blosc2.SChunk.SChunk.method" to "SChunk.method".


FrancescAlted (approved):
This looks good!  I have spotted a couple of minor things (mainly docs and comments) that I'll fix after merging.

PR #70, opened 2022-11-08
os.popen2 is deprecated → subprocess.run

DimitriPapadopoulos:
See:
https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module


FrancescAlted:
Thanks!

PR #75, opened 2022-11-10
LICENSE → LICENSE.txt

DimitriPapadopoulos:
* For consistency with other Blosc projects, `python-blosc` and `c-blosc2`
* For consistency with Python packaging documentation:
  https://packaging.python.org//guides/distributing-packages-using-setuptools/


FrancescAlted:
LGTM

PR #76, opened 2022-11-10
Apply DeepSource.io suggestions

DimitriPapadopoulos:
* Unnecessary `None` provided as default


DimitriPapadopoulos:
While `dict.get(..., None)` and `dict.get(...)` should be identical, `dict.pop(..., None)` and `dict.pop(...)` certainly are not. I need to re-visit this merge request.

DimitriPapadopoulos:
I had mistakenly changed an instance of `dict.pop()` instead of `dict.get()`.

FrancescAlted:
Thanks!

PR #79, opened 2022-12-06
Support user defined codecs and filters

martaiborra:


blosc2/SChunk.py
```diff
@@ -104,7 +104,8 @@ def __init__(self, chunksize=None, data=None, **kwargs):
         >>> schunk = blosc2.SChunk(**storage)
         """
         # Check only allowed kwarg are passed
-        allowed_kwargs = all(kwarg in ["urlpath", "contiguous", "cparams", "dparams", "schunk", "mode"]
+        allowed_kwargs = all(kwarg in ["urlpath", "contiguous", "cparams", "dparams", "schunk",
+                                       "mode", "_is_view"]
                              for kwarg in kwargs.keys())
         if not allowed_kwargs:
             raise ValueError("Only `urlpath`, `contiguous`, `cparams`, `dparams` and `mode` "
```
FrancescAlted, line 111: `schunk` is not among the allowed kwargs.  Is that intended?  It would not be better to isolate the not allowed keyword and raise `f"Not allowed kwarg : {not_allowed_kwarg}"`?

blosc2/blosc2_ext.pyx
```diff
@@ -498,11 +540,34 @@ def get_blocksize():
     """
     return blosc1_get_blocksize()
 
+cdef _check_cparams(blosc2_cparams *cparams):
+    if cparams.nthreads > 1:
+        if BLOSC2_USER_REGISTERED_CODECS_START <= cparams.compcode <= BLOSC2_USER_REGISTERED_CODECS_STOP:
+            raise ValueError("Cannot use multi-threading with user defined codecs")
```
FrancescAlted, line 546: This could be supported in the future, could you replace "Cannot use" -> "Cannot use (yet)"

blosc2/blosc2_ext.pyx
```diff
@@ -498,11 +540,34 @@ def get_blocksize():
...
+        elif any([BLOSC2_USER_REGISTERED_FILTERS_START <= filter <= BLOSC2_USER_REGISTERED_FILTERS_STOP
+                  for filter in cparams.filters]):
+            raise ValueError("Cannot use multi-threading with user defined filters")
```
FrancescAlted, line 549: Ditto

blosc2/blosc2_ext.pyx
```diff
@@ -498,11 +540,34 @@ def get_blocksize():
...
+        elif cparams.prefilter != NULL:
+            raise ValueError("`nthreads` must be 1 when a prefilter is set")
```
FrancescAlted, line 551: "... is set" -> "... is set.  In the future this limitation could be dropped tough."

blosc2/blosc2_ext.pyx
```diff
@@ -498,11 +540,34 @@ def get_blocksize():
...
+
+cdef _check_dparams(blosc2_dparams* dparams, blosc2_cparams* cparams=NULL):
+    if cparams == NULL:
+        return
+    if dparams.nthreads > 1:
+        if BLOSC2_USER_REGISTERED_CODECS_START <= cparams.compcode <= BLOSC2_USER_REGISTERED_CODECS_STOP:
+            raise ValueError("Cannot use multi-threading with user defined codecs")
```
FrancescAlted, line 558: Ditto

blosc2/blosc2_ext.pyx
```diff
@@ -498,11 +540,34 @@ def get_blocksize():
...
+        elif any([BLOSC2_USER_REGISTERED_FILTERS_START <= filter <= BLOSC2_USER_REGISTERED_FILTERS_STOP
+                  for filter in cparams.filters]):
+            raise ValueError("Cannot use multi-threading with user defined filters")
```
FrancescAlted, line 561: Ditto

blosc2/blosc2_ext.pyx
```diff
@@ -498,11 +540,34 @@ def get_blocksize():
...
+        elif dparams.postfilter != NULL:
+            raise ValueError("`nthreads` must be 1 when a postfilter is set")
```
FrancescAlted, line 563: Ditto

blosc2/core.py
```diff
@@ -1204,3 +1204,130 @@ def schunk_from_cframe(cframe, copy=False):
...
+    version: int
+        Codec version. Default is 1.
+
+    Returns
+    -------
+    out: None
+
+    Notes
+    -----
+    * Cannot use multi-threading when using an user defined codec.
```
FrancescAlted, line 1235: "Cannot use" -> "Cannot (yet) use"

blosc2/core.py
```diff
@@ -1204,3 +1204,130 @@ def schunk_from_cframe(cframe, copy=False):
...
+        This will receive an input as a ndarray of dtype uint8, an output to fill
+        as a ndarray of dtype uint8, the filter meta and the `SChunk` instance.
+
+    Returns
+    -------
+    out: None
+
+    Notes
+    -----
+    * Cannot use multi-threading when using an user defined filter.
```
FrancescAlted, line 1300: Ditto

FrancescAlted (commented):
Great job!  I just have some suggestions (mainly derived from the fact that multi-threading might be enabled in the future, hopefully ;-).

Also, can you double check that valgrind is not reporting any leaks when running examples or tests of the new codec/filter functionality?  Thanks!

FrancescAlted (commented):
After pondering about this, I think now that we should not try to send the message that removing the single-threading limitation in the current implementation is going to be fixed anytime soon (rather, this should be done in the context of a project).  So please consider my suggestions of changing the docstrings in this regard as void.

blosc2/SChunk.py
martaiborra: Fixed in 65dcb6bf8dad3deae5d0c0fcb5cf9b95ec17da3f.

FrancescAlted (approved):

PR #80, opened 2022-12-29
Fix typos found by codespell

DimitriPapadopoulos:


RELEASE_NOTES.md
```diff
@@ -42,7 +42,7 @@ XXX version-specific blurb XXX
 
 * Lots of small improvements in the style, consistency and other glitches in the code.  Thanks to Dimitri Papadopoulos for hist attention to detail.
 
-* No need to compile C-Blosc2 tests, benchs or fuzzers.  Compilation time is much shorter now.
+* No need to compile C-Blosc2 tests, benches or fuzzers.  Compilation time is much shorter now.
```
FrancescAlted, line 45: This should actually translated by benchmarks (or benchs for short), but that can be fixed later.

DimitriPapadopoulos: Indeed, I thought so, but also thought the plural of the abbreviation would still be benches. Replacing by benchmarks as you did is the best solution.

Issue #81, opened 2023-01-09
Test failure: UnboundLocalError: local variable 'f' referenced before assignment


ArchangeGabriel:
While trying to build on Arch Linux, I get this during the test suite:
```
________________________ test_comp_info[Codec.BLOSCLZ] _________________________

codec = <Codec.BLOSCLZ: 0>

    @pytest.mark.parametrize("codec", list(blosc2.Codec))
    def test_comp_info(codec):
        blosc2.compressor_list()
        blosc2.clib_info(codec)
        blosc2.set_compressor(codec)
        assert codec.name.lower() == blosc2.get_compressor()
>       blosc2.print_versions()

tests/test_comp_info.py:21: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
blosc2/core.py:1044: in print_versions
    distro = os_release_pretty_name()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def os_release_pretty_name():
        for p in ("/etc/os-release", "/usr/lib/os-release"):
            try:
                f = open(p)
                for line in f:
                    name, _, value = line.rstrip().partition("=")
                    if name == "PRETTY_NAME":
                        if len(value) >= 2 and value[0] in "\"'" and value[0] == value[-1]:
                            value = value[1:-1]
                        return value
            except OSError:
                pass
            finally:
>               f.close()
E               UnboundLocalError: local variable 'f' referenced before assignment

blosc2/core.py:1025: UnboundLocalError
----------------------------- Captured stdout call -----------------------------
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
python-blosc2 version: 2.0.0
Blosc version: 2.6.1 ($Date:: 2022-12-15 #$)
Compressors available: ['blosclz', 'lz4', 'lz4hc', 'zlib', 'zstd']
Compressor library versions:
  BLOSCLZ: 2.5.1
  LZ4: 1.9.4
  LZ4HC: 1.9.4
  ZLIB: 1.2.13
  ZSTD: 1.5.2
[...]

PR #82, opened 2023-01-09
Fix os_release_pretty_name when os-release does not exists

ArchangeGabriel:
Fixes GH-81.


FrancescAlted:
Thanks @ArchangeGabriel !

PR #91, opened 2023-02-14
Automatic guesses for chunk and block partitions

FrancescAlted:
This should allow to make `chunks` and `blocks` in constructors optional.


martaiborra (approved):
LGTM

PR #93, opened 2023-02-20
Merge b2nd (former Caterva)

FrancescAlted:
This includes new b2nd (previously Caterva) functionality.  A new NDArray class has been introduced for leveraging multidimensionality.


FrancescAlted:
This is ready to go.  Let's start the dance.

PR #97, opened 2023-03-01
Optimise PNG compression

DimitriPapadopoulos:
Use `optipng` to reduce size of PNG files between 30% to 75%.

For smaller distributions, but mostly for faster download of images from online documentation.

As a complement to c847ad476a20971611b8b7789be38701f11d4786.

By the way, I see a discrepancy between [examples/images/ucodecs-filters/forward.png](https://github.com/Blosc/python-blosc2/blob/main/examples/images/ucodecs-filters/forward.png) which has transparent background and [examples/images/ucodecs-filters/backward.png](https://github.com/Blosc/python-blosc2/blob/main/examples/images/ucodecs-filters/backward.png) which has white background.


FrancescAlted:
Thanks @DimitriPapadopoulos, that's a great reduction :-). Regarding the incosistency in transparent background, that is unfortunate.  I think we prefer transparent.  Do you know if there a way to convert the background in transparent for an existing PNG?

DimitriPapadopoulos:
It's probably best to re-export the SVG source file [examples/images/ucodecs-filters/backward.svg](https://github.com/Blosc/python-blosc2/blob/main/examples/images/ucodecs-filters/backward.svg) to PNG than modifying the PNG file [examples/images/ucodecs-filters/backward.png – especially around anti-aliased fonts. I don't know which software has been used to export SVG files. Inkspace?

FrancescAlted:
Yes, it is inkscape.

PR #98, opened 2023-03-01
Add global registered codecs and filters

martaiborra:


FrancescAlted (approved):
LGTM.  When using blosc2.print_versions() we don't get them for plugins, but that is minor and can be fixed later on.

PR #99, opened 2023-03-02
Transparent background in PNG files exported from SVG

DimitriPapadopoulos:
The exported PNG images are slightly different because I have exported from Linux instead of Windows.

You might want to re-export from Windows to obtain the exact same result – especially for font rendering and size.


FrancescAlted:
Great.  And Linux rendering is fine (actually, it is difficult for me to notice the difference :-).  Thanks!

PR #101, opened 2023-03-09
Bytedelta

FrancescAlted:
This allows to use the [new bytedelta filter](https://github.com/Blosc/c-blosc2/pull/456) from C-Blosc2.

Some scripts for downloading real data and transcoding it has been added too.


blosc2/blosc2_ext.pyx
```diff
@@ -2036,6 +2043,11 @@ cdef b2nd_context_t* create_b2nd_context(shape, chunks, blocks, dtype, kwargs):
     # This is used only in constructors, dtype will always have NumPy format
     dtype = np.dtype(dtype)
     typesize = dtype.itemsize
+    if 'cparams' in kwargs:
+        if 'typesize' not in kwargs['cparams']:
+            kwargs['cparams']['typesize'] = typesize
+    else:
+        kwargs['cparams'] = {'typesize': typesize}
```
martaiborra, line 2050: This won't override the type size if passed. We should changed the doc of some ndarray constructors.

martaiborra (commented):
Other than a minor thing, LGTM.

blosc2/blosc2_ext.pyx
FrancescAlted: Good point.  Fixed in d5095eb

PR #106, opened 2023-05-03
Workaround for ndindex issues with wheels on Win32

FrancescAlted:
Bu using conda-forge package for ndindex, GHA tests seems to work again.


FrancescAlted:
This has been reported in:
https://github.com/Quansight-Labs/ndindex/issues/156

PR #119, opened 2023-06-13
Iterchunks info

FrancescAlted:
This is a new chunk iterator optimized for retrieving metadata from chunks: cratio, whether they contain special values, and if so, which is the special value.

This is handy for iterating over datasets with a large number of chunks that are special; e.g. if some chunk is all made of zeros, it can be skipped during a reduction operation.  This can provide large speedups in e.g. sparse arrays (or arrays with lots of NaNs, or repeated values).

### DIFF
--- a/blosc2/ndarray.py
+++ b/blosc2/ndarray.py
@@ -179,20 +179,21 @@ class NDArray(blosc2_ext.NDArray):
             A namedtuple with the following fields:
             nchunk: the index of the chunk.
             coords: the coordinates where the chunk starts.
+            cratio: the compression ratio of the chunk.
             special: the special value enum of the chunk; if 0, the chunk is not special.
             repeated_value: the repeated value for the chunk; if not SpecialValue.VALUE, it is None.
             lazychunk: a buffer with the complete lazy chunk.
         """
         ChunkInfoNDArray = namedtuple(
-            "ChunkInfoNDArray", ["nchunk", "coords", "special", "repeated_value", "lazychunk"]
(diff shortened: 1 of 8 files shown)


blosc2/ndarray.py
```diff
@@ -168,6 +169,32 @@ def __setitem__(self, key, value):
...
+            coords: the coordinates where the chunk starts.
+            cratio: the compression ratio of the chunk.
+            special: the special value enum of the chunk; if 0, the chunk is not special.
+            repeated_value: the repeated value for the chunk; if not SpecialValue.VALUE, it is None.
+            lazychunk: a buffer with the complete lazy chunk.
+        """
+        ChunkInfoNDArray = namedtuple(
+            "ChunkInfoNDArray", ["nchunk", "coords", "cratio", "special", "repeated_value", "lazychunk"]
+        )
+        chunks_idx = np.array(self.shape) // np.array(self.chunks)
```
martaiborra, line 190: This should be` np.array(self.ext_shape) // np.array(self.ext_chunks)` in case there's padding.

martaiborra (commented):
Other than supporting for shapes and chunk shapes that are not divisible, LGTM.

blosc2/ndarray.py
FrancescAlted: Thanks.  Fixed in https://github.com/Blosc/python-blosc2/pull/119/commits/bdd16148c7b666db5cea11c323d6cf199d16d4fc

PR #120, opened 2023-06-15
Now register_codec can be used to register dynamic plugins

jdavid:
If both encoder and decoder are None, then the codec must be a dynamic plugin, installed as a wheel.


FrancescAlted (approved):
register_codec is already in the top_level section, but we can merge this and fix that later.

martaiborra (approved):
LGTM

PR #124, opened 2023-07-08
Initialize codec and filter params

jdavid:
Fixes segfault with a plugin that uses these params, such as blosc2_openhtj2k


martaiborra (approved):
LGTM!

PR #127, opened 2023-08-01
Check multithreading only for python plugin functions

martaiborra:
This lets use multithreading for user-defined plugins when the corresponding functions are not Python functions.


FrancescAlted (changes requested):
Looks good to me.  If not there yet, I'd suggest to add a couple of tests, one when using a Python codec with nthreads > 1, and other for a Python filter.

FrancescAlted (approved):
Great.  Thanks!

PR #136, opened 2023-09-29
Fix error in 0830349

DimitriPapadopoulos:
After 0830349:
```
f-string: invalid conversion character: expected 's', 'r', or 'a'
```


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #138, opened 2023-09-30
Avoid duplicate iteration over the same dict

DimitriPapadopoulos:
It was probably thought as an optimisation in #79, but it's really a duplicate iteration.


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #139, opened 2023-10-04
Add tuner parameter in cparams

martaiborra:
With this PR now you can specify the tuner to use like you would do with the codec.

### DIFF
--- a/blosc2/core.py
+++ b/blosc2/core.py
@@ -1230,12 +1230,12 @@ def compute_chunks_blocks(shape, chunks=None, blocks=None, dtype=np.uint8, **kwa
                     cparams2["filters"][i] = blosc2.Filter.NOFILTER
         else:
             cparams2 = cparams
-        # Do not apply specific tuner
+        # Force STUNE to get a hint on the blocksize
         aux_tuner = cparams2["tuner"] if "tuner" in cparams2 else blosc2.Tuner.STUNE
         cparams2["tuner"] = blosc2.Tuner.STUNE
         src = blosc2.compress2(np.zeros(nitems, dtype=f"V{itemsize}"), **cparams2)
-        cparams2["tuner"] = aux_tuner
         _, _, blocksize = blosc2.get_cbuffer_sizes(src)
+        cparams2["tuner"] = aux_tuner
         # Starting point for the guess
         if chunks is None:
             blocks = [1 if shape[i] == 1 else 2 for i in range(len(shape))]
--- a/blosc2/core.py
+++ b/blosc2/core.py
@@ -1230,7 +1230,11 @@ def compute_chunks_blocks(shape, chunks=None, blocks=None, dtype=np.uint8, **kwa
                     cparams2["filters"][i] = blosc2.Filter.NOFILTER
         else:
             cparams2 = cparams
+        # Do not apply specific tuner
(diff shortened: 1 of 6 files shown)


blosc2/core.py
```diff
@@ -1230,7 +1230,11 @@ def compute_chunks_blocks(shape, chunks=None, blocks=None, dtype=np.uint8, **kwa
                     cparams2["filters"][i] = blosc2.Filter.NOFILTER
         else:
             cparams2 = cparams
+        # Do not apply specific tuner
+        aux_tuner = cparams2["tuner"] if "tuner" in cparams2 else blosc2.Tuner.STUNE
+        cparams2["tuner"] = blosc2.Tuner.STUNE
```
FrancescAlted, line 1235: Can you add a comment above like "# Force STUNE to get a hint on the blocksize"

blosc2/core.py
```diff
@@ -1230,7 +1230,11 @@ def compute_chunks_blocks(shape, chunks=None, blocks=None, dtype=np.uint8, **kwa
...
         src = blosc2.compress2(np.zeros(nitems, dtype=f"V{itemsize}"), **cparams2)
+        cparams2["tuner"] = aux_tuner
```
FrancescAlted, line 1237: I think moving this line one position below would increase readability

doc/reference/autofiles/top_level/blosc2.Tuner.rst
```diff
@@ -0,0 +1,13 @@
...
+.. currentmodule:: blosc2
+
+.. autoclass:: Tuner
+   
+   .. rubric:: Attributes
+
+   .. autosummary::
+   
+      ~Tuner.STUNE
+      ~Tuner.BTUNE
```
FrancescAlted, line 13: I'd add some docstrings here:

STUNE: A 'simple' tuner.  This is the default in the Blosc2 library

BTUNE: A more sophisticated tuner that can select different codecs/filters for different chunks.  More info: https://btune.blosc.org

FrancescAlted (changes requested):
Apart of the comments, it would be nice to add some example for BTUNE (would require to have `blosc2-btune` installed).

FrancescAlted:
> Apart of the comments, it would be nice to add some example for BTUNE (would require to have `blosc2-btune` installed).

After thinking twice, the best would be to do the example in the `blosc2-btune` project.

FrancescAlted:
> > Apart of the comments, it would be nice to add some example for BTUNE (would require to have `blosc2-btune` installed).
> 
> After thinking twice, the best would be to do the example in the `blosc2-btune` project.

Or perhaps in both, why not .

blosc2/core.py
martaiborra: Done in 5d6795a2f591109e1b70a1bca2cbeb6c11844cb5

blosc2/core.py
martaiborra: Fixed in 5d6795a2f591109e1b70a1bca2cbeb6c11844cb5

doc/reference/autofiles/top_level/blosc2.Tuner.rst
martaiborra: Fixed in 5d6795a2f591109e1b70a1bca2cbeb6c11844cb5

FrancescAlted (approved):
LGTM

PR #145, opened 2023-11-01
Add optional offset to open function

ivilata:
The new optional `offset` argument to `blosc2.schunk.open` allows providing an offset for Blosc2 super-chunks stored inside of other container files (like HDF5). Doc updates and a simple unit test are provided.

### DIFF
--- a/tests/test_open.py
+++ b/tests/test_open.py
@@ -83,3 +83,39 @@ def test_open(contiguous, urlpath, cparams, dparams, nchunks, chunk_nitems, dtyp
 def test_open_fake():
     with pytest.raises(FileNotFoundError):
         _ = blosc2.open("none.b2nd")
+
+
+@pytest.mark.parametrize("offset", [0, 42])
+@pytest.mark.parametrize("urlpath", ["schunk.b2frame"])
+def test_open_offset(offset, urlpath):
+    urlpath_temp = urlpath + ".temp"
+
+    blosc2.remove_urlpath(urlpath)
+    blosc2.remove_urlpath(urlpath_temp)
+
+    # Create a temporary file with data.
+    data = np.arange(100)
+    schunk_temp = blosc2.SChunk(data=data, urlpath=urlpath_temp)
+    del schunk_temp
+    # Create the final file with the temporary data after "offset" bytes.
+    with open(urlpath, "wb") as schunk_file:
+        schunk_temp_data = None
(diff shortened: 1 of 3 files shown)


tests/test_open.py
```diff
@@ -83,3 +83,39 @@ def test_open(contiguous, urlpath, cparams, dparams, nchunks, chunk_nitems, dtyp
...
+def test_open_offset(offset, urlpath):
+    urlpath_temp = urlpath + ".temp"
+
+    blosc2.remove_urlpath(urlpath)
+    blosc2.remove_urlpath(urlpath_temp)
+
+    # Create a temporary file with data.
+    data = np.arange(100)
+    schunk_temp = blosc2.SChunk(data=data, urlpath=urlpath_temp)
+    del schunk_temp
```
FrancescAlted, line 99: Maybe an idiom that you can use here is `_ = blosc2.SChunk(data=data, urlpath=urlpath_temp)` that would remove the need of the additional `del`?

tests/test_open.py
```diff
@@ -83,3 +83,39 @@ def test_open(contiguous, urlpath, cparams, dparams, nchunks, chunk_nitems, dtyp
...
+        with open(urlpath_temp, "rb") as schunk_temp_file:
+            schunk_temp_data = schunk_temp_file.read()
+        schunk_file.seek(offset)
+        schunk_file.write(schunk_temp_data)
+    blosc2.remove_urlpath(urlpath_temp)
+
+    schunk_open = blosc2.open(urlpath, "r", offset=offset)
+    schunk_data = schunk_open[:]
+    del schunk_open
+    assert np.array_equal(schunk_data, data.tobytes())
```
FrancescAlted, line 112: The lines above can be put more succinctly as:

```
schunk = blosc2.open(urlpath, "r", offset=offset)
assert np.array_equal(schunk[:], data.tobytes())
```

But quite cosmetic indeed :-)

FrancescAlted (approved):
I have added a couple of small suggestions, but besides that, this looks great to me.

tests/test_open.py
ivilata: Thanks! Better than that, I just did away with unneeded references in commit a6e8e15. :wink:

tests/test_open.py
ivilata: Thanks! See comment below about unneeded references.

PR #152, opened 2023-12-09
`black`/`isort` → `ruff format`

### LINKED ISSUE #151: `black`/`isort` → `ruff format`
In order to reduce the number of dependencies, I suggest using ruff instead of black and isort.

See for example https://github.com/pypa/setuptools/pull/4125#issuecomment-1822842090 and https://github.com/jaraco/skeleton/pull/99.

DimitriPapadopoulos:
Fixes #151.

Supersedes #150.


FrancescAlted:
That's a great addition. Thanks @DimitriPapadopoulos !

PR #170, opened 2024-03-19
Replace `if` expression with `min()`

DimitriPapadopoulos:


FrancescAlted:
LGTM.  Thanks @DimitriPapadopoulos !

PR #173, opened 2024-03-21
First introduction of pow operator

omaech:
This is still work in progress

### DIFF
--- a/blosc2/lazyexpr.py
+++ b/blosc2/lazyexpr.py
@@ -203,6 +203,10 @@ class LazyExpr:
     def __rpow__(self, value):
         return self.update_expr(new_op=(value, "**", self))
 
+    def __ipow__(self, value):
+        return self.update_expr(new_op=(self, "**", value))
+
+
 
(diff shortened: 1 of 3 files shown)


FrancescAlted:
That's pretty good.  Now, you have to make the test working when using `NDArray` objects, and not just `LazyExpr` instances.

FrancescAlted:
Cool.  Now, it would be nice if you can add support for `__iadd__`, `__isub__`, `__imul__` and `__itruediv__` for *both* `NDArray` and `LazyExpr`.  Add tests for both of these `NDArray` and `LazyExpr` objects too.

omaech:
I did the work in the previus commit. Would that be good enough?

FrancescAlted:
That's good!  Now, as there are much more fixtures, and to prevent an explosion in tests, could you fusion the new tests for `__iXXX__` special functions in a single one?

FrancescAlted:
Let's merge this now and continue improving later.

PR #176, opened 2024-03-25
Fix vlmeta.to_dict not honoring tuple encoding

ivilata:
All other vlmeta encoding/decoding code uses `blosc2_ext.encode/decode_tuple` to preserve tuple objects, but `vlmeta.to_dict()` in the extension was an exception:

```
>>> import blosc2
>>> a = blosc2.empty(shape=(10,10))
>>> a.schunk.vlmeta['l'] = [1,2,3]
>>> a.schunk.vlmeta['t'] = (1,2,3)
>>> a.schunk.vlmeta['l']
[1, 2, 3]
>>> a.schunk.vlmeta['t']
(1, 2, 3)
>>> a.schunk.vlmeta.to_dict()
{b'l': [1, 2, 3], b't': ['__tuple__', 1, 2, 3]}
```

This fixes the issue and performs some minor extra testing for list and tuple encoding.

### DIFF
--- a/blosc2/blosc2_ext.pyx
+++ b/blosc2/blosc2_ext.pyx
@@ -1649,7 +1649,7 @@ cdef class vlmeta:
             raise RuntimeError
         res = {}
         for i in range(rc):
-            res[names[i]] = unpackb(self.get_vlmeta(names[i]))
+            res[names[i]] = unpackb(self.get_vlmeta(names[i]), list_hook=decode_tuple)
         return res
 
 
--- a/tests/test_vlmeta.py
+++ b/tests/test_vlmeta.py
@@ -57,6 +57,7 @@ def test_schunk(contiguous, urlpath, nbytes, cparams, dparams, nchunks):
         assert nchunks_ == (i + 1)
 
     add(schunk)
+    to_dict(schunk)
     iter(schunk)
     delete(schunk)
     clear(schunk)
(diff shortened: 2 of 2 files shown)


ivilata:
As a side comment, I wasn't able to figure out why this ad hoc tuple encoding is performed by Blosc2 (which deviates from standard msgpack), but this fix does at least keep things consistent.

FrancescAlted:
> As a side comment, I wasn't able to figure out why this ad hoc tuple encoding is performed by Blosc2 (which deviates from standard msgpack), but this fix does at least keep things consistent.

That's a good point indeed.  What about removing tuple support entirely and use a msgpack list to represent it?  I'd go for it now before python-blosc2 is too much spread.

ivilata:
Umm, I did a couple of tests to find out where the special encoding was used. A quick hack making `blosc2_ext.encode/decode_tuple` the identity function resulted in tensor unit test errors like:

```
FAILED tests/test_tensor.py::test_pack_array2[1000000.0-int64] - TypeError: can not serialize 'tuple' object
```

A more elaborate test removing all uses of `default=blosc2_ext.encode_tuple, strict_types=True` in `msgpack.packb` and `list_hook=blosc2_ext.decode_tuple` in `unpackb` resulted in errors like:

```
FAILED tests/test_tensor.py::test_pack_array2_struct[100000-i4,i4] - TypeError: Field elements must be 2- or 3-tuples, got '['f0', '<i4']'
```

Converting `dtype` to a tuple in case it's a list in `blosc2.core._unpack_tensor` didn't work either, some files seem to have the encoded `__tuple__` item:

```
FAILED tests/test_tensor.py::test_pack_array2_struct[100000-i4,i4] - TypeError: Tuple must have size 2, but has size 3
```

So it is used indeed at some places, so I'd prefer if  @martaiborra chimed in. Maybe a good way to go for the moment would be to add some comment in the code indicating why this is needed and then merge the PR, as it seems like removing the special encoding is not possible/desirable now.

FrancescAlted (approved):
LGTM

PR #181, opened 2024-04-01
Avoid multiple imports of numpy

DimitriPapadopoulos:
Unless I am missing something, there's no need to duplicate:
https://github.com/Blosc/python-blosc2/blob/4fcc352e76ec554c934e7fab959af6a1238bff63/blosc2/core.py#L15


FrancescAlted:
Agreed.  Thanks @DimitriPapadopoulos !

PR #182, opened 2024-04-01
Avoid cyclic import

DimitriPapadopoulos:
Trying to fix cyclic import.


FrancescAlted:
Is that still a thing, or it can be closed?

DimitriPapadopoulos:
Probably not really a thing :smile:

PR #183, opened 2024-04-01
Use min/max builtins

DimitriPapadopoulos:


FrancescAlted:
I'm not sure I like this sort of replacements.  In my mind, the existing code is very clear already.  Which rational is under your suggestion?

DimitriPapadopoulos:
I agree.

I just tried to address the [deepsource.com](https://deepsource.com/) issues that made the most sense:
https://app.deepsource.com/gh/DimitriPapadopoulos/python-blosc2
However, I do not find this linter makes useful suggestions - less than _ruff_ any way.

PR #195, opened 2024-05-20
Reductions (and broadcast support)

FrancescAlted:
Most common reductions ("sum", "mean", "std", "min", "max", "all", "any") are supported now in LazyExpr.

Bonus: broadcasting is supported as well. Although some rough edges would require some polishing, this should come later, when people starts to experiment with it.


martaiborra (approved):
LGTM!

PR #198, opened 2024-05-26
Add Python bindings for memory-mapped files

### LINKED ISSUE #188: Partially reading NDArray slow via NFS share
Hey there,

first and foremost I would like to express my gratitude for this amazing work! It's super fast + addresses exactly the pain points that have been plaguing us for a long time.

We are currently exploring using blosc2 as a new interface for storing and loading preprocessed data for training AI-based segmentation models. Our [nnU-Net](https://github.com/MIC-DKFZ/nnUNet) repository is quite popular and optimizing its pipeline would benefit an entire community. Especially now that datasets are getting larger, the ability to efficiently store and partially read files is becoming increasingly valuable.

Let me briefly explain what it is we are doing. During training, we need to read the training data over and over again. Due to the large images that are typical for our domain (~ around 500x500x500 pixels, sometimes 20000x20000 or 1000x1000x1000.), we need to take smaller crops that the GPU can actually handle. These crops are usually around 192x192x192 or smaller. To maximize the variability of data we are presenting, we are not computing these crops offline, but rather partially read them from the larger images on the fly while the training runs. 

Currently, this is implemented with numpy memmaps. Preprocessed images are stored on some fast file system (either local SSDs or fast network shares). We use uncompressed numpy arrays (np.save) that we open with `np.load([...], mmap_mode='r')`. This is fast (provided I/O keeps up) and uses basically no CPU at all. But it requires a lot of storage and ... well ... I/O does not longer keep up.

The blosc2 NDarray offers exactly what we need to alleviate our problems: fast, partial reading of compressed arrays. An initial prototype that I implemented does exactly that. CPU usage barely increases over the numpy memmap baseline, storage requirement is drastically reduced (compression ratio ~3-4) and I/O as reported by iotop is down as well. At least when running the code on my local workstation.

On our compute cluster, when using the local SSDs of the our compute nodes, blosc2 performance is as expected. The problem lies in the NFS share. If the data is located there, performance tanks. I have a hard time pinpointing what the problem is, especially because reading memmaps with higher I/O requirements runs smoothly. 
When using blosc2, CPUs are barely used and almost all processes report that they are waiting for I/O. But when checking the network activity I see that we are only reading ~15MB/s and (curiously) also transmitting ~15MB/s? These numbers are well below what our NFS file shares are capable of (>1.5GB/s). Do you have any idea why this could be? 
Could it be that blosc2 sends a large amount of small read requests? But: Even if each chunk was read individually, this should still be fewer reads than we currently have to do with numpy memmaps as there are fewer chunks in a crop than there are contiguous rows to extract from an uncompressed array. Is there something I am not seeing?

Here is some additional information in the hope that it helps narrowing down the problem:

This is how we save arrays:
```python
        cparams = {
            'codec': blosc2.Codec.ZSTD,
            'clevel': 8,
            'nthreads': 8
        }
        blosc2.asarray(np.ascontiguousarray(data), urlpath=output_filename_truncated + '.b2nd', chunks=chunks, blocks=blocks, cparams=cparams)
```

This is how we open the files to partially read/uncompress the arrays:
```python
data = blosc2.open(urlpath=data_b2nd_file, mode='r', dparams={'nthreads': 1})
crop = data[crop_slice]
```

Info for a representative training image
> type    : NDArray
> shape   : (1, 690, 563, 563)
> chunks  : (1, 80, 64, 64)
> blocks  : (1, 16, 16, 16)
> dtype   : float32
> cratio  : 4.83
> cparams : {'blocksize': 16384,
>  'clevel': 8,
>  'codec': <Codec.ZSTD: 5>,
>  'codec_meta': 0,
>  'filters': [<Filter.NOFILTER: 0>,
>              <Filter.NOFILTER: 0>,
>              <Filter.NOFILTER: 0>,
>              <Filter.NOFILTER: 0>,
>              <Filter.NOFILTER: 0>,
>              <Filter.SHUFFLE: 1>],
>  'filters_meta': [0, 0, 0, 0, 0, 0],
>  'nthreads': 8,
>  'splitmode': <SplitMode.ALWAYS_SPLIT: 1>,
>  'typesize': 4,
>  'use_dict': 0}
> dparams : {'nthreads': 1}

When using numpy arrays, I observed that the OS is caching read operations. For smaller datasets, this substantially reduces the amount of reads over time, reducing the I/O burden. I have not observed read caching with blosc2. Is there a specific reason for that?

Thank you for you support and for providing this wonderful package! If there is anything else you need, please let me know!
Best,
Fabian

JanSellner:
This PR updates the Python bindings to bring memory-mapped files to python-blosc2 (fixes #188).

There are basically two new parameters for the user: `mmap_mode` and `initial_mapping_size` with the same meaning as in the C interface. I added them to `blosc2.open` and `blosc2.SChunk` because these seem to be the two entrypoints to supply an `urlpath`.

This PR requires [https://github.com/Blosc/c-blosc2/pull/611](https://github.com/Blosc/c-blosc2/pull/611) to be merged. I will update the submodule when the PR gets merged into c-blosc2.


blosc2/blosc2_ext.pyx
```diff
@@ -877,21 +913,33 @@ cdef class SChunk:
...
             self.schunk = <blosc2_schunk *> PyCapsule_GetPointer(_schunk, <char *> "blosc2_schunk*")
-            if mode == "w" and urlpath is not None:
+            if self.mode == "w" and urlpath is not None:
                 blosc2.remove_urlpath(urlpath)
                 self.schunk = blosc2_schunk_new(self.schunk.storage)
             return
 
         if kwargs is not None:
-            if mode == "w":
+            if self.mode == "w":
```
FrancescAlted, line 940: Well spotted

blosc2/blosc2_ext.pyx
```diff
@@ -1819,10 +1869,32 @@ def meta_keys(self):
     return keys
 
 
-def open(urlpath, mode, offset, **kwargs):
+def open(urlpath, mode, mmap_mode, initial_mapping_size, offset, **kwargs):
```
FrancescAlted, line 1872: But that important, but as mmap is a detail of the backend, I'd pack them in `kwargs`.  Agreed?

blosc2/schunk.py
```diff
@@ -929,7 +987,9 @@ def __dealloc__(self):
         super().__dealloc__()
 
 
-def open(urlpath, mode="a", offset=0, **kwargs):
+@_inherit_doc_parameter(SChunk.__init__, "mmap_mode:", {r"\* - 'w\+'[^*]+": ""})
+@_inherit_doc_parameter(SChunk.__init__, "initial_mapping_size:", {r"r\+ w\+, or c": "r+ or c"})
+def open(urlpath, mode="a", mmap_mode=None, initial_mapping_size=None, offset=0, **kwargs):
```
FrancescAlted, line 992: The positions of the mmap-related params are actually breaking the API. I'd move them into **kwargs and unpack them internally.  Using the `kwargs` mechanism allows for more robust API (e.g. future I/O plugins).

tests/test_mmap.py
```diff
@@ -0,0 +1,83 @@
...
+import numpy as np
+import pytest
+
+import blosc2
+
+
+@pytest.mark.parametrize("initial_mapping_size", [None, 1000])
+def test_open_offset(tmp_path, monkeypatch, capfd, initial_mapping_size):
+    monkeypatch.setenv("BLOSC_INFO", "true")
+    expected_mapping_size = 1073741824 if initial_mapping_size is None else initial_mapping_size
```
FrancescAlted, line 12: Probably 2**30 reads better than 1073741824, but this is in the eye of the beholder.

FrancescAlted (changes requested):
Well done.  I have some comments, specially on API changes.

Also, it would be nice if you can add an example (`examples/` dir), and optionally, but always nice to see, some benchmark comparing the new mmap IO backed with the default one.

Thanks!

blosc2/blosc2_ext.pyx
JanSellner: Yes agreed, putting the new arguments into kwargs makes much more sense.

tests/test_mmap.py
JanSellner: I also think `2**30` reads much nicer :-)

blosc2/blosc2_ext.pyx
JanSellner: :-)

JanSellner:
@FrancescAlted I changed the interface and added examples as well as a benchmark in `bench/io.py`.

Here some benchmarking results on Linux:
```
# python -m bench.io --io-type io_file --blosc-mode schunk
Time for writing the data with io_file: 0.874s
Time for reading the data with io_file in sequential order: 0.236s
Time for reading the data with io_file in random order: 0.269s
# python -m bench.io --io-type io_mmap --blosc-mode schunk
Time for writing the data with io_mmap: 0.876s
Time for reading the data with io_mmap in sequential order: 0.123s
Time for reading the data with io_mmap in random order: 0.129s
# python -m bench.io --io-type io_file --blosc-mode ndarray
Time for writing the data with io_file: 1.017s
Time for reading the data with io_file in sequential order: 0.176s
Time for reading the data with io_file in random order: 0.188s
# python -m bench.io --io-type io_mmap --blosc-mode ndarray
Time for writing the data with io_mmap: 1.232s
Time for reading the data with io_mmap in sequential order: 0.170s
Time for reading the data with io_mmap in random order: 0.151s
```

When writing or reading chunks to or from a schunk, it looks similar to the benchmark in c-blosc2: write performance is comparable while reading can be significantly faster. However, I noticed that the picture looks differently when using the ndarray interface. First, it is slower for both I/O interfaces (probably because of some metadata?) and secondly, writing with mmap is slower while reading is faster or equal. I don't know where the write slowdown is coming from especially since the chunk size settings are similar in both cases.

Unfortunately, I also found a very nasty bug when writing mmap files as I observed occasional bus errors in a non reproducible manner when running the benchmark. The problem is that I am passing the `mmap_mode` argument from Python to C via `kwargs["mmap_mode"].encode("utf-8")`. This is also done similarly for other parameters like the `urlpath`. However, the problem is that the resulting string of `encode("utf-8")` is coupled with the corresponding object in Python and only exists for the lifetime of the object as managed by Python. It seems to be available for the open function but not necessarily at a later point when writing some data to the file. The problem is that I used this mode to truncate the file to the target size:
```c
    if (strcmp(mmap_file->mode, "c") != 0) {
      int rc = ftruncate(mmap_file->fd, new_size);
      if (rc < 0) {
        BLOSC_TRACE_ERROR("Cannot extend the file size to %" PRId64 " bytes (error: %s).", new_size, strerror(errno));
        return 0;
      }
    }
```
Since the memory for the `mmap_mode` string may not be available, this comparison is flawed. I think the bus errors are happening when, by incident, the location of `mmap_file->mode` contains the c character because in those cases this condition avoids the `ftruncate` and this can yield bus errors.
Unfortunately, I have to change the C code again to fix this so that the mode parameter is only used during the open call.

FrancescAlted:
> @FrancescAlted I changed the interface and added examples as well as a benchmark in `bench/io.py`.
> 
> Here some benchmarking results on Linux:
> 
> ```
> # python -m bench.io --io-type io_file --blosc-mode schunk
> Time for writing the data with io_file: 0.874s
> Time for reading the data with io_file in sequential order: 0.236s
> Time for reading the data with io_file in random order: 0.269s
> # python -m bench.io --io-type io_mmap --blosc-mode schunk
> Time for writing the data with io_mmap: 0.876s
> Time for reading the data with io_mmap in sequential order: 0.123s
> Time for reading the data with io_mmap in random order: 0.129s
> # python -m bench.io --io-type io_file --blosc-mode ndarray
> Time for writing the data with io_file: 1.017s
> Time for reading the data with io_file in sequential order: 0.176s
> Time for reading the data with io_file in random order: 0.188s
> # python -m bench.io --io-type io_mmap --blosc-mode ndarray
> Time for writing the data with io_mmap: 1.232s
> Time for reading the data with io_mmap in sequential order: 0.170s
> Time for reading the data with io_mmap in random order: 0.151s
> ```
> 
> When writing or reading chunks to or from a schunk, it looks similar to the benchmark in c-blosc2: write performance is comparable while reading can be significantly faster. However, I noticed that the picture looks differently when using the ndarray interface. First, it is slower for both I/O interfaces (probably because of some metadata?) and secondly, writing with mmap is slower while reading is faster or equal. I don't know where the write slowdown is coming from especially since the chunk size settings are similar in both cases.

Well, the NDArray is just a regular SChunk with additional metadata.  Having said that, the chunks and blocks for the 2 internal partitions are computed automatically, and that may affect performance significantly.

Having said that, the fact that the NDArray interface with `io_file` is faster in your benchmark than the SChunk makes me suspicious about it; you know, benchmarking is a very tricky thing.  I'd prefer to wait until you have fixed the remaining issues and have a try on a couple of my boxes.  Once we have more statistics, we can continue discussing :-)

> 
> Unfortunately, I also found a very nasty bug when writing mmap files as I observed occasional bus errors in a non reproducible manner when running the benchmark. The problem is that I am passing the `mmap_mode` argument from Python to C via `kwargs["mmap_mode"].encode("utf-8")`. This is also done similarly for other parameters like the `urlpath`. However, the problem is that the resulting string of `encode("utf-8")` is coupled with the corresponding object in Python and only exists for the lifetime of the object as managed by Python. It seems to be available for the open function but not necessarily at a later point when writing some data to the file. The problem is that I used this mode to truncate the file to the target size:
> 
> ```c
>     if (strcmp(mmap_file->mode, "c") != 0) {
>       int rc = ftruncate(mmap_file->fd, new_size);
>       if (rc < 0) {
>         BLOSC_TRACE_ERROR("Cannot extend the file size to %" PRId64 " bytes (error: %s).", new_size, strerror(errno));
>         return 0;
>       }
>     }
> ```
> 
> Since the memory for the `mmap_mode` string may not be available, this comparison is flawed. I think the bus errors are happening when, by incident, the location of `mmap_file->mode` contains the c character because in those cases this condition avoids the `ftruncate` and this can yield bus errors. Unfortunately, I have to change the C code again to fix this so that the mode parameter is only used during the open call.

Yes, your diagnostic seems correct to me (that bit us quite a few times too ;-).  No worries, take your time.

FrancescAlted (changes requested):
I have a suggestion for the decorator for docstrings that you have introduced.  Tell me what you think.

FrancescAlted:
Ok, I did some benchmarking with the new version.  But first, I found the root on why the ndarray interface had better performance.  It turns out that SChunk and NDArray have different algorithms for computing the size of the first level partition (blocks), mainly due to dimensionality requirements of the second.  So I ended with a new benchmark (attached) that is explicit on the blocksize, and it also disables compression (random data do not compress well, and only wastes CPU here).

[new-io.py.txt](https://github.com/user-attachments/files/15533504/new-io.py.txt)

With that, here there are the results on my Linux box (intel 13900K):

```
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_file --blosc-mode schunk
Time for writing the data with io_file: 0.313s (3.120 GB/s)
Time for reading the data with io_file in sequential order: 0.150s (6.495 GB/s)
Time for reading the data with io_file in random order: 0.139s (7.004 GB/s)
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_file --blosc-mode ndarray
Time for writing the data with io_file: 0.439s (2.224 GB/s)
Time for reading the data with io_file in sequential order: 0.123s (7.924 GB/s)
Time for reading the data with io_file in random order: 0.121s (8.073 GB/s)

(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_mmap --blosc-mode schunk
Time for writing the data with io_mmap: 0.462s (2.113 GB/s)
Time for reading the data with io_mmap in sequential order: 0.034s (28.787 GB/s)
Time for reading the data with io_mmap in random order: 0.035s (27.922 GB/s)
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_mmap --blosc-mode ndarray
Time for writing the data with io_mmap: 0.838s (1.165 GB/s)
Time for reading the data with io_mmap in sequential order: 0.125s (7.836 GB/s)
Time for reading the data with io_mmap in random order: 0.119s (8.212 GB/s)
```

So, with the regular io_file driver, one can see that performance of the schunk and ndarray is closer now (still a bit different, though).

However, for the io_mmap driver, schunk is *much* faster than ndarray.  This was kind of what I was expecting, because the double partitioning for multidimensions is intrinsically more difficult to realize than a linear one.  But for this kind of benchmark, where you read a chunk at a time, I realized that a faster path could be done and, as this should be a common pattern access  [I implemented it](https://github.com/Blosc/c-blosc2/commit/a874872e3db80d67f083422c880663f843e485a0).  Also, I was able [to accelerate the `NDArray.__getitem__`](https://github.com/Blosc/python-blosc2/commit/5c7e1052185f6965c0def0a2ac56cb8f5954ccfb) special method for this use case too.

With the above improvements, here there are the results:

```
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_file --blosc-mode schunk
Time for writing the data with io_file: 0.308s (3.167 GB/s)
Time for reading the data with io_file in sequential order: 0.122s (7.992 GB/s)
Time for reading the data with io_file in random order: 0.129s (7.544 GB/s)
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_file --blosc-mode ndarray
Time for writing the data with io_file: 0.440s (2.217 GB/s)
Time for reading the data with io_file in sequential order: 0.043s (22.460 GB/s)
Time for reading the data with io_file in random order: 0.045s (21.736 GB/s)

(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_mmap --blosc-mode schunk
Time for writing the data with io_mmap: 0.460s (2.122 GB/s)
Time for reading the data with io_mmap in sequential order: 0.036s (27.313 GB/s)
Time for reading the data with io_mmap in random order: 0.035s (27.552 GB/s)
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_mmap --blosc-mode ndarray
Time for writing the data with io_mmap: 0.815s (1.198 GB/s)
Time for reading the data with io_mmap in sequential order: 0.037s (26.697 GB/s)
Time for reading the data with io_mmap in random order: 0.039s (24.913 GB/s)
```

In this case, performance for schunk and ndarray interfaces for mmap are very similar (and close to the memory bandwidth of my system, so we are good there).

With that, I think that the PR is good to go (bar the comment about docstrings and decorators in my latest review).

FrancescAlted:
For completeness, here there are the results on my Macbook Air (Apple M1):

```
francesc@Francescs-MacBook-Air ~/b/python-blosc2 (JanSellner-mmap)> python -m bench.io --io-type io_file --blosc-mode schunk
Time for writing the data with io_file: 0.899s (1.086 GB/s)
Time for reading the data with io_file in sequential order: 0.197s (4.948 GB/s)
Time for reading the data with io_file in random order: 0.201s (4.862 GB/s)
francesc@Francescs-MacBook-Air ~/b/python-blosc2 (JanSellner-mmap)> python -m bench.io --io-type io_file --blosc-mode ndarray
Time for writing the data with io_file: 0.738s (1.324 GB/s)
Time for reading the data with io_file in sequential order: 0.134s (7.309 GB/s)
Time for reading the data with io_file in random order: 0.133s (7.316 GB/s)

francesc@Francescs-MacBook-Air ~/b/python-blosc2 (JanSellner-mmap)> python -m bench.io --io-type io_mmap --blosc-mode schunk
Time for writing the data with io_mmap: 0.361s (2.702 GB/s)
Time for reading the data with io_mmap in sequential order: 0.068s (14.451 GB/s)
Time for reading the data with io_mmap in random order: 0.068s (14.413 GB/s)
francesc@Francescs-MacBook-Air ~/b/python-blosc2 (JanSellner-mmap)> python -m bench.io --io-type io_mmap --blosc-mode ndarray
Time for writing the data with io_mmap: 0.851s (1.148 GB/s)
Time for reading the data with io_mmap in sequential order: 0.069s (14.193 GB/s)
Time for reading the data with io_mmap in random order: 0.069s (14.226 GB/s)
```

This time, write results with io_mmap are better (compared to io_file) than with my Linux box. But for me, strongest selling point of mmap is read performance, so I am glad to see the important boost on that part :-)

blosc2/schunk.py
```diff
@@ -947,13 +1043,15 @@ def open(urlpath, mode="a", offset=0, **kwargs):
...
     kwargs: dict, optional
         Keyword arguments supported:
-        cparams: dict
-            A dictionary with the compression parameters, which are the same that can be
-            used in the :func:`~blosc2.compress2` function.
-            Typesize and blocksize cannot be changed.
-        dparams: dict
-            A dictionary with the decompression parameters, which are the same that can
-            be used in the :func:`~blosc2.decompress2` function.
+            mmap_mode:
```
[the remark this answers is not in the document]
JanSellner: This actually fixes an indentation issue. With the current indentation, sphinx does not typeset the cparams correctly: [https://www.blosc.org/python-blosc2/reference/autofiles/top_level/blosc2.open.html#blosc2-open](https://www.blosc.org/python-blosc2/reference/autofiles/top_level/blosc2.open.html#blosc2-open)

blosc2/schunk.py
```diff
@@ -929,6 +1023,8 @@ def __dealloc__(self):
         super().__dealloc__()
 
 
+@_inherit_doc_parameter(SChunk.__init__, "mmap_mode:", {r"\* - 'w\+'[^*]+": ""})
+@_inherit_doc_parameter(SChunk.__init__, "initial_mapping_size:", {r"r\+ w\+, or c": "r+ or c"})
```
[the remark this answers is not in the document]
JanSellner: I first thought about making a link from `open` to `SChunk.__init__` which would definitively be the easiest option^^

However, I am not much a fan of hyperlinks to other functions for reading the parameter documentation of the _current_ function. This always breaks the flow for the user since, usually, one wants to know what the current function is capable of without leaving the page. For some functions, there is even another level of indirection, e.g. if I want to know the possible arguments of `blosc2.asarray`, I first have to follow the documentation to `blosc2.empty` and then again to the documentation to `SChunk.__init__`.

This decorator is an intend to fix this problem without copying documentation manually. Additionally, I also alter the documentation here (the `w+` argument is removed because it cannot be used with `blosc2.open`).

I am not saying that the decorator is the best option btw. I think this is an inherent disadvantage of sphinx that it has no built-in mechanism for these use cases. A sphinx extensions might be better suited but I have no experience there and don't know whether this is even possible. On the other hand, with the decorator, `help(blosc2.open)` works properly.

What do you think?

JanSellner:
@FrancescAlted Thanks a lot for looking so detailed at the benchmark and great that you implemented a faster path for sequential reading (which I can also confirm on my box). This is really cool and could also be beneficial for our spectral video data when we need all the frames. If I understand this condition
```c
if (fast_path && slice_size == array->chunknitems) {
```
correctly, the fast path can only be used if we read everything but cannot be used if we need a contiguous series of chunks (e.g., frames 10 to 100)?

I totally agree that reading is the main selling point of mmap which is also the more important metric I think. At least for AI use cases where you prepare your data once and then use it continuously during training. The only exception might be writing to NFS shares (or maybe other network drives) where I could imagine that also mmap writing is faster. But this is mainly guessing as this very hard to reproduce not to mention benchmarking. I also updated the documentation to reflect this. In any case, it is great that users will soon have the option to try this out :-)

FrancescAlted:
> @FrancescAlted Thanks a lot for looking so detailed at the benchmark and great that you implemented a faster path for sequential reading (which I can also confirm on my box). This is really cool and could also be beneficial for our spectral video data when we need all the frames. If I understand this condition
> 
> ```c
> if (fast_path && slice_size == array->chunknitems) {
> ```
> 
> correctly, the fast path can only be used if we read everything but cannot be used if we need a contiguous series of chunks (e.g., frames 10 to 100)?

Yes, the fast path is only for reading a *single* chunk.  Accelerating for a contiguous series of chunks requires more work (that you are welcome to pursue if that is a big need for you).

> I totally agree that reading is the main selling point of mmap which is also the more important metric I think. At least for AI use cases where you prepare your data once and then use it continuously during training. The only exception might be writing to NFS shares (or maybe other network drives) where I could imagine that also mmap writing is faster. But this is mainly guessing as this very hard to reproduce not to mention benchmarking. I also updated the documentation to reflect this. In any case, it is great that users will soon have the option to try this out :-)

Ok. You are right, creating a new I/O backend is not easy exactly because of the reasons you mention.  Well, let's hope that the rough edges can be ironed out and we can get rid of the 'beta' label soon :-)

FrancescAlted:
I have been reflecting on the new fast path for reading single, well-behaved, chunks, and I thought that a similar fast path can be provided for writing too, and I [did that here](https://github.com/Blosc/c-blosc2/commit/addcc195e1eb5629a1b9e6d8938df4dc06bf910e).  My next thought has been to use the new [fastpath for accelerating `blosc2.asarray()`](https://github.com/Blosc/python-blosc2/commit/de8696329d7e578eaea23451ed270a4c73fed989).

In the end that worked quite well, at least for the `io_file` method, and on my Linux box:

```
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_file --blosc-mode schunk
Time for writing the data with io_file: 0.316 s (3.090 GB/s)
Time for reading the data with io_file in sequential order: 0.145 s (6.721 GB/s)
Time for reading the data with io_file in random order: 0.149 s (6.558 GB/s)
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_file --blosc-mode ndarray
Time for writing the data with io_file: 0.350 s (2.793 GB/s)
Time for reading the data with io_file in sequential order: 0.046 s (21.389 GB/s)
Time for reading the data with io_file in random order: 0.044 s (21.945 GB/s)

(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_mmap --blosc-mode schunk
Time for writing the data with io_mmap: 0.465 s (2.102 GB/s)
Time for reading the data with io_mmap in sequential order: 0.035 s (28.023 GB/s)
Time for reading the data with io_mmap in random order: 0.035 s (27.963 GB/s)
(python-blosc2) faltet@beast:/mnt/ssd0/faltet$ PYTHONPATH=~/blosc/python-blosc2 python ~/blosc/python-blosc2/bench/io.py --io-type io_mmap --blosc-mode ndarray
Time for writing the data with io_mmap: 1.128 s (0.865 GB/s)
Time for reading the data with io_mmap in sequential order: 0.038 s (25.961 GB/s)
Time for reading the data with io_mmap in random order: 0.035 s (27.574 GB/s)
```

So, write performance has increased a 1.3x for `io_file`, which is quite good.  Unfortunately, performance for mmap has *decreased* almost a 1.4x (I was expecting the other way around actually, so I am a bit sad).  I suppose write patterns are important for optimizing write performance, and the new algorithm should be using one that is not too good for mmap.

At any rate, as the new fast path is using a blocked algorithm, it also provides the advantage of ingesting large arrays that do not fit in-memory, so I'm afraid it is here to stay.  At very least, perhaps it will offer a good test bench for optimizing that part for the io_mmap driver.

BTW, my latest commits broke your PR.  Sorry about that.

blosc2/schunk.py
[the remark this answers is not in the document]
FrancescAlted: Yeah. Ok, let's do it.  I don't entirely like the fact that in most parts of docs we use an indirection, while in others, the docs will be just duplicated; but as this should go mostly unnoticed by users, that's fine, I suppose.

FrancescAlted:
And for reference, here there the results of the bench.io on my MacBook Air (M1):

```
> python -m bench.io --io-type io_file --blosc-mode schunk
Time for writing the data with io_file: 0.622 s (1.569 GB/s)
Time for reading the data with io_file in sequential order: 0.186 s (5.251 GB/s)
Time for reading the data with io_file in random order: 0.174 s (5.599 GB/s)
> python -m bench.io --io-type io_file --blosc-mode ndarray
Time for writing the data with io_file: 0.703 s (1.389 GB/s)
Time for reading the data with io_file in sequential order: 0.133 s (7.327 GB/s)
Time for reading the data with io_file in random order: 0.133 s (7.325 GB/s)

> python -m bench.io --io-type io_mmap --blosc-mode schunk
Time for writing the data with io_mmap: 0.355 s (2.750 GB/s)
Time for reading the data with io_mmap in sequential order: 0.066 s (14.723 GB/s)
Time for reading the data with io_mmap in random order: 0.066 s (14.710 GB/s)
francesc@Francescs-MacBook-Air ~/b/python-blosc2 (JanSellner-mmap)> python -m bench.io --io-type io_mmap --blosc-mode ndarray
Time for writing the data with io_mmap: 1.229 s (0.794 GB/s)
Time for reading the data with io_mmap in sequential order: 0.069 s (14.143 GB/s)
Time for reading the data with io_mmap in random order: 0.069 s (14.224 GB/s)
```

Again, one can see that, for `io_file` the new fastpath for `blosc2.asarray()` has improved speed a bit, whereas for `io_mmap`, it is significantly slower.

FrancescAlted (approved):

FrancescAlted:
As we have other changes in queue, and to prevent breaking this PR once again, I'm going to proceed with the merge now.

FrancescAlted:
Thanks @JanSellner !

JanSellner:
@FrancescAlted Thanks a lot for merging and your continuing effort on faster paths. Actually, the fast path for writing sounds even more generally applicable since it should perfectly fit to the common use case of writing whole, existing arrays at once.

I did some further investigation on the mmap writing issue with some callgrind runs to find out the cause for the slowdowns. I built the python extension via `python setup.py build_ext --inplace --build-type "RelWithDebInfo"` and then compared the output for mmap and io_file (only the top functions are shown below):

Note: I disabled reading for the benchmark and wrapped the relevant lines in
```python
            valgrind.callgrind_start_instrumentation()
            t0 = time()
            blosc2.asarray(array, chunks=self.chunks, blocks=self.blocks,
                           cparams=self.cparams, mode="w",
                           mmap_mode=self.mmap_mode_write, urlpath=self.urlpath)
            t1 = time()
            valgrind.callgrind_stop_instrumentation()
```
with the help of the [`valgrind` Python](https://pypi.org/project/valgrind/) package.

**mmap:**
```bash
$ valgrind --tool=callgrind --instr-atstart=no --collect-jumps=yes --dump-instr=yes python -m bench.io --io-type io_mmap --blosc-mode ndarray --runs 1
$ callgrind_annotate callgrind.out.48829.1

--------------------------------------------------------------------------------
Ir
--------------------------------------------------------------------------------
2,835,587,326 (100.0%)  PROGRAM TOTALS

--------------------------------------------------------------------------------
Ir                      file:function
--------------------------------------------------------------------------------
1,285,903,006 (45.35%)  ./string/../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:__memcpy_avx_unaligned_erms [/usr/lib/x86_64-linux-gnu/libc.so.6]
1,179,738,054 (41.60%)  ./string/../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:__memset_chk_avx2_unaligned_erms [/usr/lib/x86_64-linux-gnu/libc.so.6]
  165,376,000 ( 5.83%)  blosc2/c-blosc2/blosc/b2nd_utils.c:copy4dim [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
   34,859,703 ( 1.23%)  blosc2/c-blosc2/blosc/fastcopy.c:copy_match [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
   25,254,759 ( 0.89%)  blosc2/c-blosc2/blosc/blosclz.c:blosclz_decompress [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
   15,791,193 ( 0.56%)  blosc2/c-blosc2/blosc/shuffle-avx2.c:unshuffle8_avx2 [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
   12,039,669 ( 0.42%)  blosc2/c-blosc2/blosc/blosclz.c:blosclz_compress [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
```

**io_file:**
```bash
valgrind --tool=callgrind --instr-atstart=no --collect-jumps=yes --dump-instr=yes python -m bench.io --io-type io_file --blosc-mode ndarray --runs 1
callgrind_annotate callgrind.out.53133.1

--------------------------------------------------------------------------------
Ir
--------------------------------------------------------------------------------
2,746,322,256 (100.0%)  PROGRAM TOTALS

--------------------------------------------------------------------------------
Ir                      file:function
--------------------------------------------------------------------------------
1,179,738,054 (42.96%)  ./string/../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:__memset_chk_avx2_unaligned_erms [/usr/lib/x86_64-linux-gnu/libc.so.6]
1,171,893,292 (42.67%)  ./string/../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:__memcpy_avx_unaligned_erms [/usr/lib/x86_64-linux-gnu/libc.so.6]
  165,376,000 ( 6.02%)  blosc2/c-blosc2/blosc/b2nd_utils.c:copy4dim [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
   34,859,703 ( 1.27%)  blosc2/c-blosc2/blosc/fastcopy.c:copy_match [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
   25,254,759 ( 0.92%)  blosc2/c-blosc2/blosc/blosclz.c:blosclz_decompress [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
   15,791,193 ( 0.57%)  blosc2/c-blosc2/blosc/shuffle-avx2.c:unshuffle8_avx2 [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
   12,039,669 ( 0.44%)  blosc2/c-blosc2/blosc/blosclz.c:blosclz_compress [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
```

As you can see, the counts for the blosc functions are identical (as expected) and the main difference comes from more `__memcpy_avx_unaligned_erms` (+114009714 ~ 9 %) and `__memset_chk_avx2_unaligned_erms` calls in the mmap case (note that the order is swapped). I can only imagine that this comes from the `memcpy(mmap_file->addr + position, ptr, n_bytes);` call in the `blosc2_stdio_mmap_write` function. This call itself does not show up in callgrind, however, I think because it is optimized away by the compiler and replaced with the call to `__memcpy_avx_unaligned_erms`.

In principle, the `memcpy` is not necessary in the mmap case because you could directly write to the target location. This would require a change of the interface (very similar to the read function). However, this would also require a change of all existing memory allocations and maybe a completely new memory management since there are also a lot of `calloc` and `realloc` situations which would require different treatment (also in the default io case if the interface is changed).

For us, this is not a big issue because we care much more about read than write performance, so I am leaving this here mainly as reference in case someone wants to tackle this problem in the future.

BTW: memset seems to take a huge amount of time in both io cases. This sounds a lot to me given that this should only be due to memory initializations. Is this expected for you?

FrancescAlted:
> @FrancescAlted Thanks a lot for merging and your continuing effort on faster paths. Actually, the fast path for writing sounds even more generally applicable since it should perfectly fit to the common use case of writing whole, existing arrays at once.
> 
> I did some further investigation on the mmap writing issue with some callgrind runs to find out the cause for the slowdowns. I built the python extension via `python setup.py build_ext --inplace --build-type "RelWithDebInfo"` and then compared the output for mmap and io_file (only the top functions are shown below):
> 
> Note: I disabled reading for the benchmark and wrapped the relevant lines in
> 
> ```python
>             valgrind.callgrind_start_instrumentation()
>             t0 = time()
>             blosc2.asarray(array, chunks=self.chunks, blocks=self.blocks,
>                            cparams=self.cparams, mode="w",
>                            mmap_mode=self.mmap_mode_write, urlpath=self.urlpath)
>             t1 = time()
>             valgrind.callgrind_stop_instrumentation()
> ```
> 
> with the help of the [`valgrind` Python](https://pypi.org/project/valgrind/) package.
> 
> **mmap:**
> 
> ```shell
> $ valgrind --tool=callgrind --instr-atstart=no --collect-jumps=yes --dump-instr=yes python -m bench.io --io-type io_mmap --blosc-mode ndarray --runs 1
> $ callgrind_annotate callgrind.out.48829.1
> 
> --------------------------------------------------------------------------------
> Ir
> --------------------------------------------------------------------------------
> 2,835,587,326 (100.0%)  PROGRAM TOTALS
> 
> --------------------------------------------------------------------------------
> Ir                      file:function
> --------------------------------------------------------------------------------
> 1,285,903,006 (45.35%)  ./string/../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:__memcpy_avx_unaligned_erms [/usr/lib/x86_64-linux-gnu/libc.so.6]
> 1,179,738,054 (41.60%)  ./string/../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:__memset_chk_avx2_unaligned_erms [/usr/lib/x86_64-linux-gnu/libc.so.6]
>   165,376,000 ( 5.83%)  blosc2/c-blosc2/blosc/b2nd_utils.c:copy4dim [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
>    34,859,703 ( 1.23%)  blosc2/c-blosc2/blosc/fastcopy.c:copy_match [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
>    25,254,759 ( 0.89%)  blosc2/c-blosc2/blosc/blosclz.c:blosclz_decompress [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
>    15,791,193 ( 0.56%)  blosc2/c-blosc2/blosc/shuffle-avx2.c:unshuffle8_avx2 [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
>    12,039,669 ( 0.42%)  blosc2/c-blosc2/blosc/blosclz.c:blosclz_compress [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
> ```
> 
> **io_file:**
> 
> ```shell
> valgrind --tool=callgrind --instr-atstart=no --collect-jumps=yes --dump-instr=yes python -m bench.io --io-type io_file --blosc-mode ndarray --runs 1
> callgrind_annotate callgrind.out.53133.1
> 
> --------------------------------------------------------------------------------
> Ir
> --------------------------------------------------------------------------------
> 2,746,322,256 (100.0%)  PROGRAM TOTALS
> 
> --------------------------------------------------------------------------------
> Ir                      file:function
> --------------------------------------------------------------------------------
> 1,179,738,054 (42.96%)  ./string/../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:__memset_chk_avx2_unaligned_erms [/usr/lib/x86_64-linux-gnu/libc.so.6]
> 1,171,893,292 (42.67%)  ./string/../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:__memcpy_avx_unaligned_erms [/usr/lib/x86_64-linux-gnu/libc.so.6]
>   165,376,000 ( 6.02%)  blosc2/c-blosc2/blosc/b2nd_utils.c:copy4dim [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
>    34,859,703 ( 1.27%)  blosc2/c-blosc2/blosc/fastcopy.c:copy_match [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
>    25,254,759 ( 0.92%)  blosc2/c-blosc2/blosc/blosclz.c:blosclz_decompress [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
>    15,791,193 ( 0.57%)  blosc2/c-blosc2/blosc/shuffle-avx2.c:unshuffle8_avx2 [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
>    12,039,669 ( 0.44%)  blosc2/c-blosc2/blosc/blosclz.c:blosclz_compress [/workspaces/python-blosc2/blosc2/blosc2_ext.cpython-312-x86_64-linux-gnu.so]
> ```
> 
> As you can see, the counts for the blosc functions are identical (as expected) and the main difference comes from more `__memcpy_avx_unaligned_erms` (+114009714 ~ 9 %) and `__memset_chk_avx2_unaligned_erms` calls in the mmap case (note that the order is swapped). I can only imagine that this comes from the `memcpy(mmap_file->addr + position, ptr, n_bytes);` call in the `blosc2_stdio_mmap_write` function. This call itself does not show up in callgrind, however, I think because it is optimized away by the compiler and replaced with the call to `__memcpy_avx_unaligned_erms`.

I am pretty sure that memset is not the issue here, as the number of calls looks identical for both file and mmap cases (memset is called a lot inside the BloscLZ codec, but I am not sure why you want to test bench.io with compression; actually it should be disabled by `clevel=0`).  But I can rather see the excess of memcpy, i.e. your  `memcpy(mmap_file->addr + position, ptr, n_bytes);` suspect, as the most probably main cause of the overhead (specially if copies are large, which should be the case).

> In principle, the `memcpy` is not necessary in the mmap case because you could directly write to the target location. This would require a change of the interface (very similar to the read function). However, this would also require a change of all existing memory allocations and maybe a completely new memory management since there are also a lot of `calloc` and `realloc` situations which would require different treatment (also in the default io case if the interface is changed).

Sounds like a big task indeed.

> For us, this is not a big issue because we care much more about read than write performance, so I am leaving this here mainly as reference in case someone wants to tackle this problem in the future.

Yeah, I have always seen mmap much more promising for reading purposes (although the gap is closing with the fast paths implemented recently).  But only for how well mmap works with NFS shares, it is great; so thanks again for tackling that.

> BTW: memset seems to take a huge amount of time in both io cases. This sounds a lot to me given that this is "only" memory initialization. Is this expected for you?

As said, if you disable compression in your bench, I am pretty sure this will go away (almost).

FWIW, we are currently working on [Caterva2](https://www.blosc.org/Caterva2/getting_started/index.html), a publisher-subscriber solution for accessing remote Blosc2 data via HTTP(S).  See an example here: https://github.com/Blosc/python-blosc2/blob/main/examples/ndarray/c2array_expr.py.  Perhaps this could be useful for you.

FrancescAlted:
@JanSellner Before releasing python-blosc2 3.0.0-b1, I think this would be a great time to enable mmap by default when using `mode="r"`.  Could you provide a PR for this?  Thanks in advance.

JanSellner:
I agree that changing the default to mmap for reading would be nice but I see some conflicts with the current API:
If we automatically use mmap for `mode="r"`, then it would not be possible anymore for the user to use the old I/O for reading. The current default for `blosc2.open` is `mode="a"` so the user has to change the arguments anyway to get read-only behavior. Using mmap for the current default would translate to `r+` -  which includes writing.

We could change the API (since the next release will be a major release, this should be ok) to mmap_mode="r" so that per default appending data would not be possible anymore but reading would be faster per default. Actually, I am not sure whether appending to an existing file is such a common use case. I would assume that in most cases the file is just read.

FrancescAlted:
I have had a chat with @martaiborra , the original author of the file `mode` support in Blosc2, and we have some suggestions regarding this discussion:

1) Try to fuse `mode` and `mmap_mode`; it is not very elegant to use different modes depending on the IO driver. Right now, the main issue that we see is the support of `mode='a'` that is not present in `mmap_mode`; can this be 'simulated' somehow when using mmap?  If so, I'd go ahead and use the modes existing in `mmap_mode` in `mode`, add 'a' to these, and finally remove `mmap_mode` entirely.

2) If 1) can be achieved, the default mode can continue being 'a', and it will use the mmap driver for doing all I/O.  The only mode that will use the file driver is going to be 'w'.  We think this will provide reasonable performance in different scenarios; e.g. when using 'a', is because people is normally opening a file for reading (writing is possible, but not the main thing to do). As 'w' is normally used to create a new file, the regular file driver will be used, providing best performance here. The rest of the modes will use the mmap driver as default.

3) When the user wants to force the use of the file or mmap driver, they will have a new parameter called `iodriver`.  We could add a new [IODRIVER enumerated after the existing ones](https://github.com/Blosc/python-blosc2/blob/main/blosc2/__init__.py#L83).  There is an existing `io` param in [storage_defaults](https://github.com/Blosc/python-blosc2/blob/main/blosc2/__init__.py#L271), but as we did not provide any other driver than `file`, very few people (if any) will be using it, so it is completely ok to rename `io` -> `iodriver` here (and in any other place where it appears that).

How does this sound @JanSellner ?

JanSellner:
Ok, this is a bit complicated and there are many thing which come into play here so first a few remarks:

1. One of the main differences between the default I/O driver and mmap is that the former is actually not using the `mode` argument of `blosc2.open` because this argument is never passed on to `blosc2_stdio_open`. The `mode` argument is only used within Python actually. This is not the case for mmap where the `mmap_mode` parameter _is_ used by `blosc2_stdio_mmap_open`.
2. I just noticed that the modes of blosc have a very special meaning which is different from the "default" meaning of those modes: the meaning of `mode='a'` in blosc is not the same as the usual meaning of the `a` mode as e.g. used by `fopen` or `open` in Python since `a` only allows to write to the end of a file but not reading: [https://stackoverflow.com/a/30566011/2762258](https://stackoverflow.com/a/30566011/2762258). The `a` mode of blosc behaves more like `a+` or `r+` (the different initial positions don't matter for blosc). Similarly, the `w` mode of blosc behaves more like a `w+` because you can also read from the file.
3. The current interface of `blosc2.open` is very similar to [`np.load`](https://numpy.org/doc/stable/reference/generated/numpy.load.html) which also has the argument `mmap_mode` to activate memory-mapped files (with the same possible mode options).
4. I would like to stress a bit the danger of having a writing mode as default for an `open` function: This allows for accidental modifications of a file in case someone (maybe an inexperienced user) is opening the file (e.g. for testing) without noticing that the default mode allows for writing and then the user adds data to it. If the file is on a shared resource, the modification affects all the users and may even stay unnoticed until someone finds the bug. I think there is a good reason that Python is using `r` as default for their `open` function. Making writes explicit can be a life-saver in these situations.

With this, first to your proposed solution: It would be possible to use the `mode` argument also for mmap and have an additional argument to select the iodriver. It would also be possible to map `a` to `r+` even though I don't think this is a very clean solution. In any case, please note the following disadvantages with this solution:

- `r+` is a writing mode so users get decreased performance per default when writing stuff via `blosc2.open`.
- On Windows, this also changes the file size to at least 1 GiB even if the file is only read (!) per default (!!).
- The user needs write access to the file even if it should only be read. This would break stuff for a lot of users I guess.

These disadvantages come mainly from point 1. and 2. above.

**Alternative solution**
Since a major release is coming up, maybe it would be a good time to clean up the mode usage of `blosc.open` and make the mode consistent to the general meaning? Then we have, r, r+ (or a+) and w+ for default I/O and r, r+, w+ and c for mmap. Further, we could switch the default for `blosc.open` to `r`, add an iodriver argument and make it default to None. The iodriver will be mmap for `r` and the default I/O for everything else while still allowing the user to change the iodriver explicitly. This gives:
- the best reading performance for the user per default.
- the best writing performance for the user per default.
- avoids the issue of point 4.
- consistent mode names in blosc.
- requires no adjustment for everyone who is using `blosc2.open` with the default arguments and only reads the data (probably the majority).
- breaks API similarity with `np.load` (point 3) which is ok I think.

Of course, this is still a breaking change and should be communicated as such. However, I think it could be worth it given the improvements in API consistency and the improved reading performance. What do you think?

FrancescAlted:
Good and sensible analysis of the problem @JanSellner .  In general, I agree with your "Alternative solution".  The only 'drawback' may be that the test suite assumes mode 'a' for many tests, and it can be a little bit annoying having to look into them and determine whether to add an explicit `mode='a'` for making things working.  But other than that, I'm +1 for this.

martaiborra:
I agree with @FrancescAlted. 
Thanks for the accurate study on this matter @JanSellner!

JanSellner:
Ok good. I can have a look at this and make a PR but it will take a bit longer because I am not available for the next two weeks.

PR #200, opened 2024-05-27
Fields in expressions

FrancescAlted:
A new `blosc2.NDField` class that represents fields of a structured `blosc2.NDArray`.  The goal is that these fields can participate as first-class citizens in expressions.

This is still WIP.


FrancescAlted:
This should be ready for review now.

martaiborra (approved):
As a preliminary version, LGTM!

PR #201, opened 2024-05-30
Add support for remote arrays

martaiborra:
Support for remote arrays as C2Array. This object can work with expressions, udf and reductions.

### DIFF
--- a/blosc2/lazyexpr.py
+++ b/blosc2/lazyexpr.py
@@ -7,6 +7,7 @@
 #######################################################################
 import copy
 import math
+import pathlib
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from enum import Enum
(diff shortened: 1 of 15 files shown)


blosc2/lazyexpr.py
```diff
@@ -402,6 +407,98 @@ def fast_eval(
...
+        else:
+            if has_padding:
+                out[slice_] = result
+            else:
+                out.schunk.update_data(nchunk, result, copy=False)
+
+    return out
+
+
+def chunks_eval(expression: str | Callable, operands: dict, **kwargs) -> blosc2.NDArray | np.ndarray:
```
FrancescAlted, line 499: This should not be necessary anymore.

FrancescAlted: Fixed in e0bc473

FrancescAlted (approved):
LGTM

Issue #204, opened 2024-06-10
Potential bug with reading existing Schunk from file?


kkotyk:
Does this line have a bug, or am I incorrectly understanding how to load an Schunk from file?
https://github.com/Blosc/python-blosc2/blob/7fac85e8035f7c7de66580f2ead60990a0ae2420/blosc2/blosc2_ext.pyx#L944

Should this line actually be?
`self.mode == "r" and urlpath is None`

FrancescAlted:
That's a good question.  May be @JanSellner can shed some light on this.

JanSellner:
Actually, the (rare) use case of opening a SChunk with `mode="r"` is currently not possible because `blosc2_schunk_new` is called in `SChunk.__init__` and this requires a non-existent file, i.e. opposite of what ``mode="r"`` requires. The tests also skip this case at the moment.

I created a PR #206 to fix this (rare) use case.

PR #206, opened 2024-06-11
Allow to create a SChunk in reading mode

JanSellner:
Related issue: #204 

With this PR, it is possible to open a schunk in reading mode (`mode="r"` or `mmap_mode="r"`) via `SChunk.__init__`. I also adapted the tests so that this use case is covered. The reading mode is now similar to `blosc2.open` but explicitly only works for schunks and not ndarrays (opening an existing ndarray via the SChunk interface probably does not make any sense).

This is now also in accordance with the documentation of `SChunk.__init__` which explicitly mention the reading modes.


blosc2/blosc2_ext.pyx
```diff
@@ -941,6 +942,8 @@ cdef class SChunk:
         if kwargs is not None:
             if self.mode == "w":
                 blosc2.remove_urlpath(urlpath)
+            elif self.mode == "r" and (urlpath is None or not os.path.exists(urlpath)):
+                raise ValueError("SChunk must already exist")
```
FrancescAlted, line 946: I have discussed this with Marta, and we have had a hard time understanding this message.  What about "When `mode` is 'r', SChunk must already exist on-disk; if not, you cannot specify `mode`".
Would that be informative enough?

blosc2/blosc2_ext.pyx
```diff
@@ -1981,6 +1993,14 @@ cdef check_schunk_params(blosc2_schunk* schunk, kwargs):
             raise ValueError("Cannot change typesize with this mode")
 
 
+cdef schunk_is_ndarray(blosc2_schunk* schunk):
+    meta1 = "b2nd"
+    meta1 = meta1.encode("utf-8") if isinstance(meta1, str) else meta1
+    meta2 = "caterva"
```
FrancescAlted, line 1999: `caterva` layer is pre-2.0.  The next release of python-blosc2 will be 3.0 (beta), so this is perfect time to get rid of this check.

tests/test_schunk.py
```diff
@@ -47,43 +47,74 @@ def test_schunk_numpy(contiguous, urlpath, mode, mmap_mode, cparams, dparams, nc
...
+    "mode_write, mode_read, mmap_mode_write, mmap_mode_read",
+    [("w", "r", None, None), (None, None, "w+", "r")],
+)
+def test_schunk_ndarray(tmp_path, mode_write, mode_read, mmap_mode_write, mmap_mode_read):
+    urlpath = tmp_path / "test.b2nd"
 
-        blosc2.remove_urlpath(urlpath)
+    data = np.arange(2 * 10, dtype="int32")
+    blosc2.asarray(data, urlpath=urlpath, mode=mode_write, mmap_mode=mmap_mode_write)
+    with pytest.raises(ValueError, match="Cannot open an ndarray as a SChunk"):
```
FrancescAlted, line 116: `ndarray` -> `NDArray` . The case is important here, most specially to distinguish from a NumPy ndarray.

blosc2/blosc2_ext.pyx
JanSellner: True, this message was not super clear^^
I expanded the message and also made it explicit whether no path was given or the file does not exist.

blosc2/blosc2_ext.pyx
JanSellner: Note: I also removed the `test_persistency_caterva` test to make this work. The file `persistency.cat` was probably created pre-2.0.

FrancescAlted (approved):
LGTM

martaiborra (approved):
LGTM!

FrancescAlted:
Thanks @JanSellner !

FrancescAlted:
FWIW in commit 466d98a I added that `SChunk.__init__()` can also be used for opening existing super-chunks (on-disk).  Also, I have made 100 cols max for docstrings (although PEP 8 recommends 72).

PR #207, opened 2024-06-13
Login support for Caterva2 context manager

ivilata:
This extends `blosc2.c2context()` with `username` and `password` parameters. When they are provided, the context manager performs a login to the Caterva2 subscriber and uses the returned authorization token for subsequent operations. Their default values may be specified via the `BLOSC_C2USERNAME` and `BLOSC_C2PASSWORD` environment variables.

Tests fixtures have been rewritten to ease enabling authentication for relevant `C2Array` tests. A new test has been added which explicitly checks opening such an array with authentication (by creating a throwaway user at the subscriber).

Some fixes to the semantics and documentation for parameter inheritance and defaults for the context manager are included too.

### DIFF
--- a/blosc2/c2array.py
+++ b/blosc2/c2array.py
@@ -109,10 +109,6 @@ def _sub_url(urlbase, path):
             else f"{urlbase}/{path}")
 
 
-def _sub_api_url(urlbase, apipath):
-    return _sub_url(urlbase, f"api/{apipath}")
-
-
 def login(username, password, urlbase):
     url = _sub_url(urlbase, "auth/jwt/login")
     creds = dict(username=username, password=password)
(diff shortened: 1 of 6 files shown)


blosc2/c2array.py
```diff
@@ -21,48 +21,83 @@
 C2SUB_URLBASE_ENVVAR = 'BLOSC_C2URLBASE'
 """Environment variable with a default Caterva2 subscriber URL base."""
 
+C2SUB_USERNAME_ENVVAR = 'BLOSC_C2USERNAME'
```
FrancescAlted, line 24: To me, I'd have a better grasp at the code if I see strings like `'BLOSC_C2USERNAME'` as-is, instead of `C2SUB_USERNAME_ENVVAR` but not a show-stopper at all.

blosc2/c2array.py
```diff
@@ -21,48 +21,83 @@
...
     Parameters
     ----------
-    urlbase: str | None
+    urlbase : str | None
         A URL base that will be used when an individual C2Array instance has
-        no subscriber URL base set.  Use ``None`` to disable the base set by a
-        previous context manager (and default to the value in the
-        ``BLOSC_C2URLBASE`` environment variable, if set).
-    auth_token: str | None
+        no subscriber URL base set.  Use the ``BLOSC_C2URLBASE`` environment
```
FrancescAlted, line 62: This is an example on why using the `C2SUB_URLBASE_ENVVAR` above is a bit pointless

FrancescAlted (approved):
I have just a minor comment; other than that, this looks great to me.

blosc2/c2array.py
ivilata: Yeah, wasn't totally convinced about how that panned out. :slightly_smiling_face:

PR #213, opened 2024-06-24
Enforce ruff/pyupgrade new rule UP031

DimitriPapadopoulos:
UP031 Use format specifiers instead of percent format


FrancescAlted:
Thanks @DimitriPapadopoulos

PR #215, opened 2024-06-26
Use `in`

DimitriPapadopoulos:


FrancescAlted (approved):
It is not very clear to me the new version reads a lot better, but let's go.

DimitriPapadopoulos:
I agree. While it's clearer for 3 equalities or more, for 2 equalities it's overkill.

PR #219, opened 2024-06-27
Unnecessary None provided as default

DimitriPapadopoulos:
`None` is the default value by default, this is redundant.


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #220, opened 2024-06-27
Do not use `==` when comparing types

DimitriPapadopoulos:
Use `is` or `isinstance` instead.


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #222, opened 2024-06-27
Do not use bare `except`

DimitriPapadopoulos:
At the very least catch `Except` instead of the default `BaseExcept`, thus not catching:
- `BaseExceptionGroup`,
- `GeneratorExit`,
- `KeyboardInterrupt`,
- `SystemExit`.


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #229, opened 2024-07-04
Enforce some ruff/tryceratops rules (TRY)

DimitriPapadopoulos:
This is a breaking change: `TypeError` is raised instead of `ValueError`.


FrancescAlted (approved):
It is a breaking change, but minimal.  Also, python-blosc2 3.0 is still beta, so a good time for this.

PR #254, opened 2024-09-11
Support filling an SChunk with a specific value. This fixes #250

### LINKED ISSUE #250: Fixing Errors with the fill_special Method for Filling an SChunk with Special Values.
When trying to use the fill_special method to fill an SChunk with a special value, I encountered an error related to the provided value. The issue was with how the special value was assigned to special_value. Initially, I tried using: special_value = blosc2.SpecialValue.VALUE. However, this assignment was incorrect because SpecialValue.VALUE was not the appropriate value and required an additional parameter to be correctly interpreted. The solution was to use a valid special value that does not require additional parameters. Instead of SpecialValue.VALUE, I used SpecialValue.ZERO, which is a predefined special value suitable for this purpose.

Additionally, the fill_special() method can be used to create arrays of zeros, NaNs, or other specific values, rather than relying on current approaches that use NumPy tricks (e.g., blosc2_ext.zeros() or blosc2_ext.full()).

martaiborra:


src/blosc2/blosc2_ext.pyx
```diff
@@ -1207,8 +1219,55 @@ cdef class SChunk:
...
+        get_chunk_repeatval(dereference(cparams), self.chunksize, chunk, chunksize, buf)
+
+        for i in range(nchunks):
+            if blosc2_schunk_append_chunk(self.schunk, <uint8_t *>chunk, True) < 0:
+                free(chunk)
+                PyBuffer_Release(buf)
+                free(buf)
+                raise RuntimeError("Error while appending the chunk")
+
+        last_nitems = nitems % self.chunkshape
```
FrancescAlted, line 1259: I'd call this variable `remainder`.

tests/test_schunk_fill.py
```diff
@@ -0,0 +1,65 @@
+#######################################################################
```
FrancescAlted, line 1: As this test builds a `SChunk`, I'd move it to `test_schunk_constructor.py`

FrancescAlted (changes requested):
Well done!  I just have a couple a suggestions.  If possible, please also make sure that the different NDArray constructors use this new way of building SChunks.

Also, it would be nice if you can attach a message in the PR with the improvements in speed for e.g. `blosc2.zeros()` and `blosc2.full()`

src/blosc2/blosc2_ext.pyx
martaiborra: Done in 296712f60e1f58169fe0f3d0766d30ff8d981e30.

tests/test_schunk_fill.py
martaiborra: Fixed in 296712f60e1f58169fe0f3d0766d30ff8d981e30.

martaiborra:
Ran a bench comparing 3 ways of filling a SChunk. The output is the following:
``` 
Time for filling the schunk with `data` argument in the constructor: 0.011s
Time for filling the schunk without passing directly the value: 0.009s
Time for filling the schunk passing directly the value to `fill_special`: 0.006s
```

FrancescAlted:
Great work!

FWIW, I simplified the benchmark (essentially letting the chunksize to be computed automatically) and pushed the size of the `schunk` (or better, the ancillary NumPy array) towards the limits of my machine (Intel i13900K, 64 GB RAM), and here are the results:

```
(python-blosc2) faltet@beast:~/blosc/python-blosc2$ python bench/fill_special.py
Filling a SChunk with 2.0 Gelements of dtype=dtype('float64')
Time with `data` argument in constructor:            1553.859 ms
Time without passing directly the value:             1580.115 ms
Time passing directly the value to `fill_special`:      0.072 ms
(python-blosc2) faltet@beast:~/blosc/python-blosc2$ python bench/fill_special.py pers
Filling a SChunk with 2.0 Gelements of dtype=dtype('float64')
Writing output to bench_fill_special.b2frame...
Time with `data` argument in constructor:            1556.498 ms
Time without passing directly the value:             1653.402 ms
Time passing directly the value to `fill_special`:     47.593 ms
```

PR #253, opened 2024-09-11
Performing Schunk-examples

omaech:

PR #255, opened 2024-09-13
Add type annotations

martaiborra:
This PR adds type annotations to the documentation. The `schunk.py` module should also be revised.


FrancescAlted (changes requested):
Looks very good so far.  Now that PR #253 is merged, please proceed with the annotations for schunk.py module.

PR #261, opened 2024-09-22
Implicit enumerate call

DimitriPapadopoulos:
Replace `range(len(expr))` with `enumerate(expr)`


FrancescAlted:
Sure. Thank you!

PR #264, opened 2024-09-23
Add CParams, DParams and Storage dataclasses

martaiborra:
This PR adds 3 data classes to pass the compression, decompression and storage parameters. The examples in the docstrings as well as the examples scripts should still be updated to use the new data classes.


FrancescAlted (approved):
Excellent work!

PR #266, opened 2024-09-25
Apply assorted ruff rules

DimitriPapadopoulos:


FrancescAlted (commented):
I don't quite understand why double quotes (") should be used instead of (') ones.  Why is this happening?  Also, the PR does not merge cleanly; can this be fixed?

Other than that, I'm fine with this.

DimitriPapadopoulos:
I have left the default value `"double"` of the [`quote-style`](https://docs.astral.sh/ruff/settings/#format_quote-style) setting of the ruff formatter. The reason is that the whole code base uses double quotes except for a few places.

I can change to `"preserve"` if you'd rather the formatter not mess with the quote style.

DimitriPapadopoulos:
I still need to fix `Python wheels / Build sdist`.

FrancescAlted (approved):
LGTM

FrancescAlted:
Ok.  I merged this.  Yeah, is not easy to deal with the quote ambiguity in Python.  Your suggested approach seems sensible to me.  Thanks!

PR #275, opened 2024-10-01
Add better support for cparams in Proxy and C2Array instances. Fixes #274.

### LINKED ISSUE #274: Proxy does not inherit the cparams/dparams properly
Here it is an example:

```python
import blosc2

cparams = blosc2.CParams(clevel=5, codec=blosc2.Codec.LZ4,
                         filters=[blosc2.Filter.BITSHUFFLE],
                         filters_meta=[0])
a = blosc2.zeros((128, 128), dtype="float64", urlpath="a.b2nd", mode="w",
                 cparams=cparams)
blosc2.remove_urlpath("proxy.b2nd")
b = blosc2.Proxy(a, urlpath="proxy.b2nd")

# And see its metadata
print(f"Codec in 'a': {a.schunk.cparams['codec']}")
print(f"Codec in 'b': {b._cache.schunk.cparams['codec']}")
print(f"Clevel in 'a': {a.schunk.cparams['clevel']}")
print(f"Clevel in 'b': {b._cache.schunk.cparams['clevel']}")
print(f"Filters in 'a': {a.schunk.cparams['filters']}")
print(f"Filters in 'b': {b._cache.schunk.cparams['filters']}")
```

and the output:

```
Codec in 'a': Codec.LZ4
Codec in 'b': Codec.ZSTD
Clevel in 'a': 5
Clevel in 'b': 1
Filters in 'a': [<Filter.BITSHUFFLE: 2>, <Filter.NOFILTER: 0>, <Filter.NOFILTER: 0>, <Filter.NOFILTER: 0>, <Filter.NOFILTER: 0>, <Filter.NOFILTER: 0>]
Filters in 'b': [<Filter.NOFILTER: 0>, <Filter.NOFILTER: 0>, <Filter.NOFILTER: 0>, <Filter.NOFILTER: 0>, <Filter.NOFILTER: 0>, <Filter.SHUFFLE: 1>]
```

FrancescAlted:


FrancescAlted:
I think this is ready to go.

@martaiborra , I still have an issue in the number of threads for compression in `cparams` of the result of an expression.  Curiously enough, this only fails (1 thread instead of several) when the whole suite is executed, but not if only e.g. `tests/ndarray/test_lazyudf.py` is selected.

My hunch is that this might be related with recent handling of making `_disable_overloaded_equal` global in 86a260b3d2b0f1e4be1eedbd922bcf4662debb00.  But I don't see this a block-stopper for the PR (unless you think differently).

src/blosc2/proxy.py
```diff
@@ -82,9 +92,10 @@ def aget_chunk(self, nchunk: int) -> bytes:
 
         Notes
         -----
-        This method is optional, and only available if the source has an async `aget_chunk` method.
+        This property is optional, and only available if the source has an async `aget_chunk` method.
```
martaiborra: property -> method

src/blosc2/proxy.py
```diff
@@ -116,6 +127,16 @@ def typesize(self) -> int:
         """
         pass
 
+    @property
+    def cparams(self) -> blosc2.CParams:
+        """
+        The compression parameters of the source.
+
+        This method is optional, and it can be overridden if the source has a different compression
+        configuration.
```
martaiborra, line 138: method -> property (?)

martaiborra (approved):
Other than minor comments, LGTM

src/blosc2/proxy.py
FrancescAlted: Fixed in 9fb5d79

src/blosc2/proxy.py
FrancescAlted: Fixed in 3cf9ac234c9d1f287e4f6ff1936f747f859b6713

Issue #276, opened 2024-10-02
_inherit_doc_parameter() decorator is not working well with Python 3.13


FrancescAlted:
I have found [_inherit_doc_parameter()](https://github.com/Blosc/python-blosc2/blob/4ac1d98e85cb76db9fb69b8a97f6a2cc32c4286c/src/blosc2/helpers.py#L4) decorator is not working correctly with Python 3.13.  For details, see errors in: https://github.com/Blosc/python-blosc2/actions/runs/11141227409/job/30961665683 .  However, the same decorator works well with 3.10, 3.11 and 3.12.  This is currently preventing creating wheels for Python 3.13.

@JanSellner any hints?

FrancescAlted:
Fixed in #277

PR #277, opened 2024-10-02
Fix doc inheritance for Python 3.13

JanSellner:
This PR fixes issue #276 

Python 3.13 automatically [strips leading whitespace in docstrings](https://docs.python.org/3.13/whatsnew/3.13.html#other-language-changes) for performance reasons which makes the `_inherit_doc_parameter()` decorator fail.

I tested the fix on Python 3.12 and Python 3.13.rc3. Generated docs look good.


FrancescAlted:
Thanks @JanSellner !

PR #297, opened 2024-11-01
Use `optipng` to reduce size of PNG files.

DimitriPapadopoulos:
For smaller package, but mostly for faster download of images in online documentation.


FrancescAlted:
Thanks @DimitriPapadopoulos

PR #303, opened 2024-11-02
Apply ruff/flake8-pytest-style rule PT006

DimitriPapadopoulos:


FrancescAlted:
I'm not sure I agree with this one, but let's play with ruff rules.

PR #305, opened 2024-11-02
Enforce Ruff-specific rules (RUF)

DimitriPapadopoulos:


FrancescAlted:
I thought `param: int = None` was explicit enough to not require `param: int | None = None` which I find kind of redundant.  Anyways...

PR #311, opened 2024-11-02
Enforce ruff/mccabe rule (C90)

### LINKED ISSUE #310: McCabe complexity
I just noticed in #309 that we did not enable a ruff equivalent to the [`max-complexity`](https://flake8.pycqa.org/en/7.1.0/user/options.html#cmdoption-flake8-max-complexity) option of flake8.

DimitriPapadopoulos:
Fixes #310.

Use the default value of the maximum McCabe complexity: 10

It looks like some functions became too complex after disabling flake8. How to handle that? Modify these functions? Disable `C901` for these functions?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,7 @@ extend-exclude = ["bench"]
 extend-select = [
     "B",
     "C4",
+    "C90",
     "I",
     "NPY",
     "PT",
--- a/src/blosc2/core.py
+++ b/src/blosc2/core.py
@@ -1279,7 +1279,7 @@ def compute_partition(nitems, maxshape, minpart=None):
     return partition
 
 
-def compute_chunks_blocks(
+def compute_chunks_blocks(  # noqa: C901
     shape: tuple[int] | list,
     chunks: tuple | list | None = None,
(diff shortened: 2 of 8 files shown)


FrancescAlted:
Interesting. For the time being, I'd prefer to ignore  `C901` on affected functions (or relax maximum McCabe complexity).

FrancescAlted:
BTW, thanks for all the careful work @DimitriPapadopoulos !  The devil is in the details, so much appreciated.

DimitriPapadopoulos:
Existing issues ignored with `# noqa: C901`.

FrancescAlted:
Sorry, same here.  Can you rebase this?

DimitriPapadopoulos:
Please merge #312 first to fix CI issues. Then let me rebase this once again.

FrancescAlted:
I am not sure I like McCabe complexity be so picky with some existing functions, but we will manage (I hope).

FrancescAlted:
Thanks @DimitriPapadopoulos !

DimitriPapadopoulos:
Or we can disable the rule. I've only added it because of the existing `.flake8` file.

PR #312, opened 2024-11-02
black and isort are not used any more (again)

DimitriPapadopoulos:
Also fix all pre-commit issues, again, so that CI tests pass.

### DIFF
--- a/examples/ndarray/reduce_and_enlarge.py
+++ b/examples/ndarray/reduce_and_enlarge.py
@@ -17,9 +17,9 @@
 import blosc2
 
 # Create arrays with specific dimensions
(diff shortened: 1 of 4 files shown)


examples/ndarray/reduce_expr_save2.py
```diff
@@ -1,9 +1,9 @@
 import blosc2
 
 # Create arrays with specific dimensions
-a = blosc2.full((2, 3, 4), 1, urlpath="a.b2nd", mode="w")    # 3D array with dimensions (2, 3, 4)
```
DimitriPapadopoulos, line 4: If you'd rather keep this as is, I could add `# fmt: off`/`# fmt: on` or `# fmt: skip` pragma comments.

FrancescAlted:
Sorry. I think I messed this up.  I'm not sure why my pre-commit did not fix this before.

PR #345, opened 2024-12-10
Config for allowing a conda build

FrancescAlted:
After this, and a new release, conda should be able to produce packages for python-blosc2.


FrancescAlted:
Note that this has changed the version from being dynamic to static.  This should be useful not only for conda builds, but for other systems as well.

PR #347, opened 2024-12-11
Fix misspellings found by codespell

DimitriPapadopoulos:
It was difficult to edit or diff these files because of the length of some lines. I had to use `sed` to fix the issues reported by `codespell`. Are these files generated from more readable files that are not part of this repository?


FrancescAlted:
Thanks @DimitriPapadopoulos !

PR #360, opened 2025-02-01
Use platformdirs for cpuinfo cache

### LINKED ISSUE #358: Caching files directly in the user's home dir
**Describe the bug**
Since 3.0 blosc2 has been dumping a json file to the user's home dir. I understand that this helps blosc start up faster. However, dumping files directly in the user's home dir is not really considered good form. [Many](https://www.reddit.com/r/linuxquestions/comments/l1b8fm/whats_the_best_way_to_clean_up_the_home_directory/) [users](https://askubuntu.com/questions/972937/how-to-get-rid-of-numerous-dotfiles-in-home-folder) are annoyed by such littering of dotfiles.

On Linux, the XDG spec is a more-or-less accepted standard for where to put data (config, cache etc). There are other standards on Mac and Windows.
I would strongly suggest using a package such as [platformdirs](https://pypi.org/project/platformdirs/) to select an appropriate location for cache files (and other files for that matter, if it becomes relevant).

Using that package (or just the xdg dirs), the cpuinfo cache file would be written to `${HOME}/.cache/python-blosc2/cpuinfo.json` on Linux (unless the user has configured another cache dir on his system).

**To Reproduce**
1. import blosc
2. file is created in home dir

**Expected behavior**

Any cache is written into the appropriate cache dir.

**Desktop (please complete the following information):**

Linux, but the issue would apply on all platforms

jonaslb:
Fixes #358. A fairly simple change to the location of caching to be in line with platform-dependent expectations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ dependencies = [
     "numexpr",
     "py-cpuinfo",
     "httpx",
+    "platformdirs",
 ]
 version = "3.0.1.dev"
 
--- a/src/blosc2/core.py
+++ b/src/blosc2/core.py
@@ -21,10 +21,11 @@ import platform
 import sys
 from dataclasses import asdict
 from functools import lru_cache
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import cpuinfo
 import numpy as np
+import platformdirs
 
 import blosc2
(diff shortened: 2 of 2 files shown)


FrancescAlted:
LGTM.  The failure in CI is unrelated, so merging.  Thanks @jonaslb !

PR #375, opened 2025-04-02
Transpose

ricardosp4:
Implemented transposed functionality for NDArrays of dimension 2.


src/blosc2/ndarray.py
```diff
@@ -3766,6 +3766,49 @@ def matmul(x1: NDArray, x2: NDArray, **kwargs: Any) -> NDArray:
...
+    Parameters
+    ----------
+    x: `NDArray`
+        The input array.
+    kwargs: Any, optional
+        Keyword arguments that are supported by the :func:`empty` constructor.
+
+    Returns
+    -------
+    out: :ref:`NDArray`
```
FrancescAlted, line 3782: Why are you using `:ref:NDArray` here and a plain `NDArray` above?  Look into other places for similar docstrings and try to be consistent.

tests/ndarray/test_matmul.py
```diff
@@ -21,11 +21,50 @@
...
+    real_part = blosc2.linspace(0, 1, shape=ashape, chunks=achunks, blocks=ablocks, dtype=dtype)
+    imag_part = blosc2.linspace(0, 1, shape=ashape, chunks=achunks, blocks=ablocks, dtype=dtype)
+    complex_matrix_a = real_part + 1j * imag_part
+    a = blosc2.asarray(complex_matrix_a)
+
+    real_part = blosc2.linspace(1, 2, shape=bshape, chunks=bchunks, blocks=bblocks, dtype=dtype)
+    imag_part = blosc2.linspace(1, 2, shape=bshape, chunks=bchunks, blocks=bblocks, dtype=dtype)
+    complex_matrix_b = real_part + 1j * imag_part
+    b = blosc2.asarray(complex_matrix_b)
+
```
FrancescAlted, line 67: In general, you should not modify the existing tests for other features (matmul) than the one that you are implementing (transpose).  But we can let this go (only for this time! ;-)

tests/ndarray/test_transpose.py
```diff
@@ -0,0 +1,103 @@
...
+    np.testing.assert_allclose(at, nat)
+
+
+@pytest.mark.parametrize(
+    "shape",
+    [
+        (3, 3, 3),
+        (12, 10, 10),
+        (10, 10, 10, 11),
+        (5, 4, 3, 2, 1, 0),
```
FrancescAlted, line 83: Be careful: the 0 dim at the end means an empty array.  Is that your intent?

FrancescAlted: Also, ndim = 1 and 2 are common use cases.  Can you add some tests for those?

FrancescAlted (commented):
I have some comments for you

tests/ndarray/test_transpose.py
ricardosp4: > Be careful: the 0 dim at the end means an empty array. Is that your intent?

I wanted to see what happens but I guess it is not very useful. I changed it to 1.

ricardosp4: > Also, ndim = 1 and 2 are common use cases. Can you add some tests for those?

That is already implemented. I'll add another vector just in case.

FrancescAlted (approved):
Looks good to me!

FrancescAlted:
Thanks @ricardosp4 !

PR #376, opened 2025-04-03
UP038 has been disabled in recent ruff releases

DimitriPapadopoulos:
https://github.com/astral-sh/ruff/releases/tag/0.10.0


DimitriPapadopoulos:
Too early... Rule just deprecated, not disabled yet.

PR #377, opened 2025-04-06
Avoid using loop variable outside the loop

DimitriPapadopoulos:
Although that's not the case here, the variable will be undefined if the list is empty.

### DIFF
--- a/tests/ndarray/test_getitem.py
+++ b/tests/ndarray/test_getitem.py
@@ -140,9 +140,10 @@ def test_iter(shape, chunks, blocks):
     npa = np.arange(int(np.prod(shape)), dtype=np.int32).reshape(shape)
     b2a = blosc2.asarray(npa, chunks=chunks, blocks=blocks)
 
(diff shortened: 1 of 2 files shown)


FrancescAlted:
LGTM.  Thanks @DimitriPapadopoulos !

FrancescAlted:
For some reason, this [made CI to crash](https://github.com/Blosc/python-blosc2/actions/runs/14308881378/job/40098904538).  It is strange that it passes the tests in the PR though...

DimitriPapadopoulos:
I fixed the issue manually. Somehow, at the moment I was writing this, I thought `len()` can iterate to get the length. Of course, that's wrong:
```python
>>> z = zip([1, 2, 3], [11, 22, 33])
>>> 
>>> len(z)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object of type 'zip' has no len()
>>> 
>>> len(list(z))
3
>>> 
```
Sorry about that.

PR #378, opened 2025-04-07
Edited NDarray slice to default to original codec/clevel for recompression 

lshaw8317:
Slice would decompress data, get slice, and recompress by default using dflt_cparams. Changed to use codec and clevel of original compressed data to recompress slice.

### DIFF
--- a/src/blosc2/ndarray.py
+++ b/src/blosc2/ndarray.py
@@ -1837,7 +1837,11 @@ class NDArray(blosc2_ext.NDArray, Operand):
         >>> print(type(c))
         <class 'blosc2.ndarray.NDArray'>
         """
-        kwargs = _check_ndarray_kwargs(**kwargs)
+        original_codec = self.cparams.codec
+        original_clevel = self.cparams.clevel
+        if "cparams" not in kwargs:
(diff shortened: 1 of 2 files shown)


src/blosc2/ndarray.py
```diff
@@ -1837,7 +1837,16 @@ def slice(self, key: int | slice | Sequence[slice], **kwargs: Any) -> NDArray:
...
+        original_codec = self.cparams.codec
+        original_clevel = self.cparams.clevel
+        original_filters = self.cparams.filters
+        if "cparams" not in kwargs:
+            kwargs["cparams"] = {
+                "codec": original_codec,
+                "clevel": original_clevel,
+                "filters": original_filters,
+            }
+        kwargs = _check_ndarray_kwargs(**kwargs)  # sets cparams to defaults
```
FrancescAlted, line 1849: You can write a more compact code with e.g.: `"codec": self.cparams.codec`

FrancescAlted (commented):
Looks good.  I just have a small suggestion.

src/blosc2/ndarray.py
lshaw8317: Ok, good idea

FrancescAlted (approved):
LGTM

PR #380, opened 2025-04-08
Fast path in NDArray.slice for slices aligned with chunks

FrancescAlted:
The title says it all


src/blosc2/ndarray.py
```diff
@@ -1029,6 +1030,92 @@ def extract_values(arr, indices: np.ndarray[np.int_], max_cache_size: int = 10)
...
+    for i, s in enumerate(key):
+        start = s.start if s.start is not None else 0
+        stop = s.stop if s.stop is not None else shape[i]
+
+        if stop > shape[i]:
+            return []
+
+        chunk_size = chunks[i]
+
+        # Ensure alignment with chunk boundaries
```
lshaw8317, line 1082: Why do you check this again within the loop?

src/blosc2/ndarray.py
```diff
@@ -1029,6 +1030,92 @@ def extract_values(arr, indices: np.ndarray[np.int_], max_cache_size: int = 10)
...
+        if start % chunk_size != 0:
```
lshaw8317: Why do you check this again in the loop?

src/blosc2/ndarray.py
```diff
@@ -1029,6 +1030,92 @@ def extract_values(arr, indices: np.ndarray[np.int_], max_cache_size: int = 10)
...
+            return []
+
+        start_idx = start // chunk_size
+        end_idx = math.ceil(stop / chunk_size)
+        start_indices.append(start_idx)
+        end_indices.append(end_idx)
+        n_chunks.append(math.ceil(shape[i] / chunk_size))
```
lshaw8317, line 1082: Do you want to do a test where the division is inexact and so math.ceil rounds up?

lshaw8317 (commented):
Two small comments

src/blosc2/ndarray.py
FrancescAlted: Yup, not really necessary.  Removed.

src/blosc2/ndarray.py
FrancescAlted: Right.  Let's use just integer division.

lshaw8317:
Looks good to me

PR #385, opened 2025-04-17
open lazyexpr don't fail if missing operands (wip)

jdavid:
To reproduce:

    blosc2.open('path-to-lazy-expression-with-missing operands')

With this change it still tries to guess the dtype and the shape, but this either fails with an error or makes the wrong guess. Because with missing operands it's imposible to correctly find out the dtype and shape, in this case dtype and shape should be None (this remains to be done).

There should be an error when trying to evaluate the expression, but not when opening it, because we may want to inspect it.


FrancescAlted:
LGTM.  Thanks @jdavid !

PR #387, opened 2025-04-18
New version for reducing memory consumption

FrancescAlted:
arange() and linspace() constructors should be better now, as they do not require temporaries.

This is still work in progress.

### DIFF
--- a/bench/ndarray/array-constructor-memray.py
+++ b/bench/ndarray/array-constructor-memray.py
@@ -1,0 +1,58 @@
+#######################################################################
+# Copyright (c) 2019-present, Blosc Development Team <blosc@blosc.org>
+# All rights reserved.
+#
+# This source code is licensed under a BSD-style license (found in the
+# LICENSE file in the root directory of this source tree)
+#######################################################################
+
+from time import time
+import os
+
+import numpy as np
+import memray
+
+import blosc2
+
+N = 100_000_000
+
+
+def info(a, t1):
+    size = a.schunk.nbytes
+    csize = a.schunk.cbytes
+    print(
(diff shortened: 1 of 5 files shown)


FrancescAlted:
With this optimization, I have been able to create a 10 TB dataset (a linspace of 850_000 x 850_000, float64; see new ``bench/ndarray/array-constructor.py`` benchmark) on disk, using less than 200 MB of RAM.  Prior to this, the RAM used exceeded 170 GB (so almost a 1000x better!).  And the time to create is also very good (< 1 hour for creating the 10 TB array); this around a 10% faster than before.  Here it is the output of the creation script:

```
*** Creating a blosc2 array with 722_500_000_000 elements (shape: (850000, 850000)) ***
Time: 3568.683 s - size: 5383.04 GB (1.51 GB/s) Storage required: 168356.28 MB (cratio: 32.7x)
        Command being timed: "python bench/ndarray/array-constructor.py"
        User time (seconds): 6226.12
        System time (seconds): 108.86
        Percent of CPU this job got: 177%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 59:28.81
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 196288
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 5
        Minor (reclaiming a frame) page faults: 55974941
        Voluntary context switches: 19291812
        Involuntary context switches: 906587
        Swaps: 0
        File system inputs: 920
        File system outputs: 344939944
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

Another advantage of this PR is that constructors need no temporary space at all, except for `blosc2.fromiter()`, that still needs a temporary (but only on-disk, and that is deleted after completion).

PR #391, opened 2025-04-29
Enable chaining of lazy expressions for logical operators

lshaw8317:
Forcibly add parenthesis on creation of lazyexpr via strings to ensure order-of-operations respected and smaller lazy expressions which compose a larger lazyexpr are executed first.
Also remove unnecessary data type modification of already-existing lazyexpr when creating new lazy expression based on this already-existing lazyexpr.


FrancescAlted (approved):
LGTM

FrancescAlted:
Thanks @lshaw8317 !

PR #398, opened 2025-05-04
Rename ruff ruleset: `TCH` → `TC`

DimitriPapadopoulos:
Ruff 0.8.0 changes the error codes for the rules in the [`flake8-type-checking`](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) category from `TCH` to `TC`. This matches changes that were made in the original `flake8-type-checking` plugin from which these rules were originally adapted.

https://astral.sh/blog/ruff-v0.8.0#new-error-codes-for-flake8-type-checking-rules

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,7 +111,7 @@ extend-select = [
     "RET",
     "RUF",
     "SIM",
-    "TCH",
+    "TC",
     "UP",
 ]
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Thanks @DimitriPapadopoulos !

Issue #397, opened 2025-05-02
get_item with lazy expressions containing reductions fails to index properly


lshaw8317:
The code below
```
a = blosc2.linspace(0, 10, num=100, shape=(10, 10))
le = blosc2.lazyexpr("sum(a, axis=0)", {'a': a})
print(le[3])
```
returns the whole array, not just the fourth element. This seems to occur with any reduction.

PR #402, opened 2025-05-06
Fix indexing for lazy expressions, and allow use of None in getitem

lshaw8317:
Solves #397. Also allows one to add arbitrary numbers of newaxes via indexing using ``None`` (as in NumPy).


FrancescAlted:
LGTM

Issue #406, opened 2025-05-13
Chaining lazy expressions


lshaw8317:
Working on this in [fixChaining](https://github.com/Blosc/python-blosc2/tree/fixChaining). Related to https://github.com/ironArray/Caterva2/issues/162. Partially fixed for operands which are LazyExpr or NDArrays. Need to support e.g. ints, as the following test fails.
```
# ##Bug 1: Chaining lazyexprs
N = 1_000
dtype = "float64"
a = blosc2.linspace(0, 1, N * N, dtype=dtype, shape=(N, N))
b = blosc2.linspace(1, 2, N * N, dtype=dtype, shape=(N, N))
c = blosc2.linspace(0, 1, N, dtype=dtype, shape=(N,))

# TODO: Eventually this test should pass
expr1 = blosc2.lazyexpr("arange(N) + b")
expr2 = blosc2.lazyexpr("a * b + 1")
expr = blosc2.lazyexpr("expr1 - expr2")
expr_final = blosc2.lazyexpr("expr * expr")
nres = (expr * expr)[:]
res = expr_final.compute()
np.testing.assert_allclose(res[:], nres)
```

PR #407, opened 2025-05-14
Fix chaining

lshaw8317:
Expand possibilities for chaining string-based lazy expressions to include main operand types (LazyExpr and NDArray). Have to incorporate other data types (which do not have shape attribute, e.g. int, float etc.) in future PR - see #406 .


FrancescAlted (approved):
Looks great!

PR #411, opened 2025-05-19
Allow use of where for string expressions and integer operands

### LINKED ISSUE #406: Chaining lazy expressions

lshaw8317:
To enable use of where in string lazyexprs, added code to correctly compile string to be passed to numexpr.
Secondly, added check in lazyexpr.dtype to allow for operands (e.g int) which do not have shape attribute. This means that chaining lazyexprs is more fully supported (see test), and so fixes #406.

### DIFF
--- a/src/blosc2/lazyexpr.py
+++ b/src/blosc2/lazyexpr.py
@@ -2789,9 +2789,9 @@ class LazyExpr(LazyArray):
                 # if new_expr has where_args, must have come from where(...) - or possibly where(where(..
                 # since 5*where, where + ...  are evaluated eagerly
                 if hasattr(new_expr, "_where_args"):
-                    st = (
-                        expression_.find("where(") + 6
-                    )  # expr always begins where( - could just set = 6 probably
+                    st = expression_.find("where(") + len(
+                        "where("
+                    )  # expr always begins where( - should have st = 6 always
                     finalexpr = ""
                     counter = 0
                     for char in expression_[st:]:  # get rid of external where(...)
--- a/src/blosc2/lazyexpr.py
+++ b/src/blosc2/lazyexpr.py
@@ -2789,7 +2789,9 @@ class LazyExpr(LazyArray):
                 # if new_expr has where_args, must have come from where(...) - or possibly where(where(..
                 # since 5*where, where + ...  are evaluated eagerly
                 if hasattr(new_expr, "_where_args"):
-                    st = expression_.find("where(") + 6  # could probably just be set = 6
+                    st = (
+                        expression_.find("where(") + 6
+                    )  # expr always begins where( - could just set = 6 probably
                     finalexpr = ""
                     counter = 0
                     for char in expression_[st:]:  # get rid of external where(...)
--- a/src/blosc2/lazyexpr.py
+++ b/src/blosc2/lazyexpr.py
@@ -2091,6 +2091,8 @@ class LazyExpr(LazyArray):
 
         operands = {
             key: np.ones(np.ones(len(value.shape), dtype=int), dtype=value.dtype)
+            if hasattr(value, "shape")
+            else value
             for key, value in self.operands.items()
         }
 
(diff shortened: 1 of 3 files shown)


tests/ndarray/test_lazyexpr_fields.py
```diff
@@ -226,6 +226,47 @@ def test_where(array_fixture):
...
+
+    # Test 3: nested wheres
+    # Test with string expression
+    expr = blosc2.lazyexpr("where(where(a2 < 0, 3, a2) > 3, 10, a1)")
+    # Test with eval
+    res = expr.compute()
+    nres = ne_evaluate("where(where(na2 < 0, 3, na2) > 3, 10, na1)")
+    np.testing.assert_allclose(res[:], nres)
+
+    # Test 4: muliplied wheres
```
FrancescAlted, line 261: Typo: muliplied -> multipled

tests/ndarray/test_lazyexpr_fields.py
```diff
@@ -226,6 +226,47 @@ def test_where(array_fixture):
...
+    # Test with string expression
+    expr = blosc2.lazyexpr("1*where(a2 < 0, 3, a2)")
```
FrancescAlted, line 263: please add spaces around operators: "1*where" -> "1 * where"

tests/ndarray/test_lazyexpr_fields.py
```diff
@@ -226,6 +226,47 @@ def test_where(array_fixture):
...
+    # Test with eval
+    res = expr.compute()
+    nres = ne_evaluate("1*where(na2 < 0, 3, na2)")
```
FrancescAlted, line 266: Ditto

src/blosc2/lazyexpr.py
```diff
@@ -2784,6 +2786,18 @@ def _new_expr(cls, expression, operands, guess, out=None, where=None, ne_args=No
                 expression_, operands_ = conserve_functions(
                     _expression, _operands, new_expr.operands | local_vars
                 )
+                # if new_expr has where_args, must have come from where(...) - or possibly where(where(..
+                # since 5*where, where + ...  are evaluated eagerly
+                if hasattr(new_expr, "_where_args"):
+                    st = expression_.find("where(") + 6  # could probably just be set = 6
```
FrancescAlted, line 2792: What it means the comment "# could probably just be set = 6" ? Or is it a stale one?

FrancescAlted (commented):
I have some small suggestions, but other than that, this looks good!

src/blosc2/lazyexpr.py
lshaw8317: It'e because the case handling means the first 6 characters of the expression will (I think) always be "where(", so using a regex is possibly unnecessary. But I think it is more readable like this. I would prefer to keep the comment for explainability

FrancescAlted (approved):
LGTM.  Thanks!

PR #414, opened 2025-05-21
Fixed bug for integer indexes

lshaw8317:
Since np.s_[1] does not return a slice object, but rather an integer, the code failed for cases such as this. Now fixed.


FrancescAlted (approved):
LGTM

PR #416, opened 2025-05-22
Fix lazyexpr o0 bug to pass test

### LINKED ISSUE #415: Demo examples/ndarray/reduce_string_expr.py has stopped working
```
> python examples/ndarray/reduce_string_expr.py
Expression: (o0 + ((((o2 ** 2) + (o2 ** 2)) + ((2 * o2) * o2)) + 1).sum() + o1.std())
Operands: {'o0': <blosc2.ndarray.NDArray object at 0x1162ccf10>, 'o1': <blosc2.ndarray.NDArray object at 0x115a4aed0>, 'o2': <blosc2.ndarray.NDArray object at 0x115dabf90>}
Reduction with Blosc2:
 [2067.73735833 2067.74238346]
Reduction with NumPy:
 [1067.73735822 1067.74238334]
Traceback (most recent call last):
  File "/Users/faltet/blosc/python-blosc2/examples/ndarray/reduce_string_expr.py", line 41, in <module>
    assert np.allclose(sum, npsum)
           ~~~~~~~~~~~^^^^^^^^^^^^
AssertionError

```

lshaw8317:
Resolves #415. The problem was in the conserve_functions function when looping over the operands of a lazyexpr which contains operands already labelled using the o- convention. Adding an illegal character to keep track of already changed operands, and then removing, solves the problem.

### DIFF
--- a/src/blosc2/lazyexpr.py
+++ b/src/blosc2/lazyexpr.py
@@ -719,11 +719,13 @@ def conserve_functions(  # noqa: C901
                         opname,
                         v,
                     ) in localop.operands.items():  # expression operands already in terms of basic operands
-                        newopname = ";" + self.update_func(v)  # add illegal character ;
+                        # add illegal character ; to track changed operands and not overwrite later
+                        newopname = ";" + self.update_func(v)
                         newexpr = re.sub(
                             rf"(?<=\s){opname}|(?<=\(){opname}", newopname, newexpr
                         )  # replace with newopname
-                    node.id = newexpr.replace(";", "")  # remove all illegal characters
+                    # remove all instances of ; as all changes completed
+                    node.id = newexpr.replace(";", "")
                 else:
                     node.id = self.update_func(localop)
             else:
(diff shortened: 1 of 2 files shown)


src/blosc2/lazyexpr.py
```diff
@@ -719,11 +719,11 @@ def visit_Name(self, node):
                         opname,
                         v,
                     ) in localop.operands.items():  # expression operands already in terms of basic operands
-                        newopname = self.update_func(v)
+                        newopname = ";" + self.update_func(v)  # add illegal character ;
```
FrancescAlted, line 722: Please explain better why are you adding an illegal character

FrancescAlted (approved):
LGTM.  I am just suggesting to add a better comment on how the fix works.

src/blosc2/lazyexpr.py
lshaw8317: Done.

PR #417, opened 2025-05-23
Enable slice lazy

lshaw8317:
Add .slice method to LazyExpr class so that it is possible to get a slice and return a blosc2.NDArray (__getitem__ returns NumPy array).


FrancescAlted:
LGTM

PR #418, opened 2025-05-24
Add support for new pandas UDF engine

### LINKED ISSUE #383: blosc2.jit support for pandas UDFs
xref https://github.com/pandas-dev/pandas/issues/61125

We discussed this informally in the past, sharing more clearly how blosc2.jit and pandas can interact.

I'm about to open a PR in pandas to support this:

```python
import pandas

def my_func(x):
    return np.sin(x * 2)

s = pandas.Series([1, 2, 3], index=list('abc'), name='sample')

# normal call executed by pandas
print(s.map(my_func))

# we let blosc2 handle this
print(s.map(my_func, engine=blosc2.jit))
```

To be able to do this, we would need blosc2 to implement a new interface. The implementation shouldn't be too complex, something like (the example ignores `skip_na` and another method `apply` for column-wise operations (function being called with the whole array, not each scalar):

```python
import numpy as np
import blosc2

# Reference base class: https://github.com/pandas-dev/pandas/blob/main/pandas/core/apply.py#L77
class Blosc2ExecutionEngine:
    @staticmethod
    def map(data, func, args, kwargs, decorator, skip_na):
        if not isinstance(data, np.ndarray):
            # we probably received a Series
            if hasattr(data, "values"):
                data = data.values
            else:
                # there is a chance that we call this with a pyarrow object in the future
                raise ValueError("blosc2.jit does not support {data.__name__}")
                
        func = decorator(func)
        result = func(data, *args, **kwargs)
        return result


blosc2.jit.__pandas_udf__ = Blosc2ExecutionEngine
```

The advantage of this approach over just decorating the function is that the whole execution loop can be jitted, not only the individual calls.

What do you think? Is this something you'd like to implement? Any feedback? It's designed in a way that you don't need to add a dependency on pandas. We aim to have Numba and Bodo supporting this same interface, and possibly others.

datapythonista:
Closes #383 

With these changes, it'll be possible to do things like this with pandas and blosc2:

```python
import pandas
import blosc2

def my_func(x):
    return np.sin(x * 2)

s = pandas.Series([1, 2, 3], index=list('abc'), name='sample')

# blosc2 will jit the function
print(s.map(my_func, engine=blosc2.jit))
```

In this PR I'm assuming that `blosc2.jit` is intended to be used with vectorized numpy operations. If that's not the case, and you want to support scalar functions and jit Python loops (see example below), let me know and we can implement this for map:

```python

@blosc2.jit
def my_func(array):
    for item in array:
        scalar_udf(item)
```

Also, `apply` is designed in pandas to call the udf for each column or row in a dataframe. I'm passing the whole array since it's a numpy array and I think operations should be vectorized, but I can make it work by columns or rows.

### DIFF
--- a/src/blosc2/proxy.py
+++ b/src/blosc2/proxy.py
@@ -676,3 +676,50 @@ def jit(func=None, *, out=None, **kwargs):  # noqa: C901
         return decorator
     else:
         return decorator(func)
+
+
+class PandasUdfEngine:
+    @staticmethod
+    def _ensure_numpy_data(data):
+        if not isinstance(data, np.ndarray):
+            try:
+                data = data.values
+            except AttributeError as err:
+                raise ValueError(
+                    "blosc2.jit received an object of type {data.__name__}, which is not supported. "
+                    "Try casting your Series or DataFrame to a NumPy dtype."
+                ) from err
+        return data
+
+    @classmethod
+    def map(cls, data, func, args, kwargs, decorator, skip_na):
+        """
+        JIT a NumPy array element-wise. In the case of Blosc2, functions are
+        expected to be vectorized NumPy operations, so the function is called
+        with the NumPy array as the function parameter, instead of calling the
+        function once for each element.
+        """
+        data = cls._ensure_numpy_data(data)
+        func = decorator(func)
+        return func(data, *args, **kwargs)
+
+    @classmethod
+    def apply(cls, data, func, args, kwargs, decorator, axis):
+        """
+        JIT a NumPy array by column or row. In the case of Blosc2, functions are
+        expected to be vectorized NumPy operations, so the function is called
+        with the NumPy array as the function parameter, instead of calling the
+        function once for each column or row.
+        """
+        data = cls._ensure_numpy_data(data)
+        func = decorator(func)
+        if data.ndim in (1, 2):
+            return func(data, *args, **kwargs)
+        else:
+            raise NotImplementedError(
+                "The blosc2 engine only supports data with with 1 or 2 dimensions. "
+                f"A NumPy array with {data.ndim} dimensions has been received."
+            )
+
+
+jit.__pandas_udf__ = PandasUdfEngine
--- a/src/blosc2/proxy.py
+++ b/src/blosc2/proxy.py
@@ -720,7 +720,7 @@ class PandasUdfEngine:
             for row_idx in range(data.shape[1]):
                 result.append(func(data[:, row_idx], *args, **kwargs))
             return np.vstack(result).transpose()
-        elif axis == (1, "columns"):
+        elif axis in (1, "columns"):
             # pandas apply(axis=1) row-wise
             result = []
             for col_idx in range(data.shape[0]):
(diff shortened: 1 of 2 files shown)


src/blosc2/proxy.py
```diff
@@ -676,3 +676,50 @@ def wrapper(*args, **func_kwargs):
...
+    def apply(cls, data, func, args, kwargs, decorator, axis):
+        """
+        JIT a NumPy array by column or row. In the case of Blosc2, functions are
+        expected to be vectorized NumPy operations, so the function is called
+        with the NumPy array as the function parameter, instead of calling the
+        function once for each column or row.
+        """
+        data = cls._ensure_numpy_data(data)
+        func = decorator(func)
+        if data.ndim in (1, 2):
```
FrancescAlted, line 716: Why this check?  I am pretty sure that Blosc2 can handle arrays up to 8 dims (can be made larger by recompiling the underlying C-Blosc2 library).  If that is not the case, this is a bug.

FrancescAlted (commented):
Good job.  I just have a small comment about dimensionality of NumPy arrays, although I don't think this is going to be too important for pandas.

datapythonista:
> BTW, do you have some preliminary benchmarks to check speed-ups? Thanks for your time!

Not at this point. I'm not sure about implementing proper benchmarking, since from the pandas side we'll just allow anyone to run the `map` and `apply`. What I would like to do is identify the best use cases for running pandas with Blosc2 (and for other known engines), and have them in the docs, with an execution time comparison. Is there any real world case you can think of where Blosc2 would be a good option to JIT a pandas Series or DataFrame?

FrancescAlted:
Yes, Blosc2 is using numexpr behind the scenes, so it will work well with any mathematical function (sin, log, exp), conditionals, where (in the sense of numpy.where); see https://www.blosc.org/python-blosc2/reference/array_operations.html for the full list.  In addition, it also supports most of the ufuncs implemented by NumPy, although the performance may not be as stellar as the ones supported by numexpr (listed above).

datapythonista:
Great, thanks for confirming. I'm wondering now if it'd be clearer for users if we just raise a NotImplementedError for map. I thought at first that it could make sense to run the vectorized function, but maybe that's confusing for users, if they try the same code with different engines. Maybe better to make engines always behave like the default engine, and raise if the exact behaviour is not supported. What do you think?

FrancescAlted:
I think `NotImplementedError` for map in Blosc2 can make sense, as I expect pretty bad performance in that mode.

datapythonista:
I updated the PR so the blosc2 engine passes to the udf the data in the way is expected. And I raise `NotImplementedError` for `map`.

FrancescAlted (approved):
LGTM

FrancescAlted:
Thank you Marc!

datapythonista:
I don't think returning a numpy array from apply is expected. But I'm not
working on pandas anymore, since there was a lot of pushback in the project
on anything I was trying to work on including this API. I guess it's a bug,
but better to open an issue in case that's a new expected behavior I don't
understand.

For the exception, the API is quite badly designed (my work on trying to
make it clearer was also often blocked and I discountinued), but I think
you need axis=1 on apply to get data row by row.

Issue #399, opened 2025-05-05
Error for indexing for lazy expressions conditions and slicing


lshaw8317:
**Describe the bug**
Indexing for lazy expressions with conditions throws an error when slicing. Related to ``chunked_eval`` ``_slice`` parameter.
In fact this problem is more serious since the ``chunked_eval`` function (called by ``expr.compute(item)`` as well) takes the slice prior to evaluating the expression, which thus gives unexpected behaviour. 

**To Reproduce**
```
N = 1000
rng = np.random.default_rng(seed=1)
it = ((-x + 1, x - 2, rng.normal()) for x in range(N))
sa = blosc2.fromiter(
    it, dtype=[("A", "i4"), ("B", "f4"), ("C", "f8")], shape=(N,), urlpath="sa-1M.b2nd", mode="w"
)
expr = sa["(A < B)"]
A = sa["A"][:]
B = sa["B"][:]
C = sa["C"][:]
temp = sa[:]
indices = (A < B)
newtemp = temp[indices]
# print(newtemp, expr[:])
print(newtemp.shape, expr[:].shape)
temp = expr.compute()
print(temp[0]) #does not throw error
print(expr[0]) #throws error - difference is get_item = True
```

**Expected behavior**
Should return one element. Possible (quick) solution - force compute of whole array and then get elements. Ideally would calculate chunkwise and stop when required number of elements reached.

lshaw8317:
Possible solution - switch to using slice_eval

PR #419, opened 2025-05-30
Make behaviour of compute consistent for slicing

lshaw8317:
As discussed in #399, there were problems for slicing lazy expressions with one where condition, since the output is in general of a different length to the input.
The .compute function when provided with a slice argument internally takes slices of the operands prior to evaluating the expression. One could change this, but I think it is better to interpret the argument as slicing operands prior to calculation. If one is interested in a general slice of the full result one can always force via expr.compute().slice(sl). 
An unfortunate consequence is that the syntax of the get_item method is a little misleading (hence why I have changed the tests).

expr.compute(sl)[:] is equal to the numpy syntax nsa1[sl][ne_evaluate(expr)[sl]].

### DIFF
--- a/src/blosc2/lazyexpr.py
+++ b/src/blosc2/lazyexpr.py
@@ -1378,7 +1378,7 @@ def slices_eval(  # noqa: C901
             for i, (c, s) in enumerate(zip(coords, chunks, strict=True))
         )
         # Check whether current slice_ intersects with _slice
-        if _slice is not None and _slice != ():  # can't use != when _slice is np.int
+        if _slice is not None and _slice is not ():  # noqa: F632  # can't use != when _slice is np.int
             # Ensure that _slice is of type slice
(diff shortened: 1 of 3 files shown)


src/blosc2/lazyexpr.py
```diff
@@ -328,7 +329,9 @@ def __getitem__(self, item: int | slice | Sequence[slice]) -> blosc2.NDArray:
         Parameters
         ----------
         item: int, slice or sequence of slices
-            The slice(s) to be retrieved. Note that step parameter is not yet honored.
+            If provided, items of the operands to be used in computation.
+            Important to note that item is used to slice the operands PRIOR to computation, not to retrieve specified
+            slices of the evaluated result.
```
FrancescAlted, line 334: Same here

FrancescAlted (approved):

PR #427, opened 2025-06-22
Add stack

lshaw8317:
Incorporate expand_dims from c-blosc2 (update github tag in CMakeLists.txt to point to latest version of c-blosc2) and add python-level blosc2.stack function to concatenate arrays along a new axis.


FrancescAlted (commented):
I have some comments.  Also, it would be nice if you can add a new `stack.py` bench (you can use `bench/ndarray/concatenate.py` as a starting point.

PR #430, opened 2025-06-29
New slices_eval_getitem() for accelerating evals with __getitem__

FrancescAlted:
The new `slices_eval_getitem()` internal function allows to evaluate things like:

```python
d = 200
shape = (d,) * 4
chunks = (d // 4,) * 4
blocks = (d // 10,) * 4
a = blosc2.ones(shape=shape, chunks=chunks, blocks=blocks)
expr = a * 2

t = time.time()
res = expr[:,:,:,1]
t3 = time.time() - t
```

more efficiently, so that t3 can be up to 10x faster than with existing code.

The solution is partial because this only works for `expr[:,:,:,1]` idiom, and not `expr.compute(item=...)`.  Also, memory handling for operand temporaries has changed: now all the operands will use the final slice prior to do the computation.  While for the case above this is (very) beneficial, this should not be the case in general.

I tried to assemble a more general solution in the new `slices_eval2()`, but it is not ready yet.  My hunch is that we could do much better code if we use the [chunking machinery](https://quansight-labs.github.io/ndindex/api/chunking.html) in the ndindex package.  Even with this, I think `slices_eval_getitem()` will always be more efficient than `slices_eval2()` for small slices, so we will probably need some heuristics to decide which one to use.

### DIFF
--- a/bench/ndarray/slice-expr.py
+++ b/bench/ndarray/slice-expr.py
@@ -2,7 +2,8 @@
 import numpy as np
 import blosc2
 import time
-
+from memory_profiler import memory_usage
+import matplotlib.pyplot as plt
 
 file = "dset-ones.b2nd"
 # a = blosc2.open(file)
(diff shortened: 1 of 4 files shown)


lshaw8317 (commented):
I will do a PR with a new test which fails with the current code, plus a fix.

PR #431, opened 2025-07-04
Refactoring slicing of lazyexprs using ndindex

lshaw8317:
Trying to simplify slicing code using NDindex. Also added functionality to squeeze function and improved case handling of indexing for blosc2 so that dummy dimensions are not introduced (thus respecting numpy responses for the same inputs).


lshaw8317:
Checked with the `slice_expr.py` benchmark and there is no notable slowdown by switching to NDindex compared to previous.

PR #435, opened 2025-07-09
Fixing reductions and slicing for lazy expressions

lshaw8317:
Now support steps for reductions. More complex slicing operations also now give correct answers. The new step_handler function could be incorporated into slice_eval to avoid numpy arrays completely and only use final_slice to get rid of dummy dimensions.


FrancescAlted (approved):
Excellent work!

Issue #436, opened 2025-07-09
Refactor `slice_eval` and other functions as necessary


lshaw8317:
Improve `slice_eval` in `lazyexpr.py` to use `step_handler` as described in #435.

lshaw8317:
Solved by PR #446

PR #446, opened 2025-08-04
Improve slice eval

lshaw8317:
Necessary bug fixes for integer slicing for reductions and refactoring slices_eval as suggested in https://github.com/Blosc/python-blosc2/issues/436. 
Had to add a couple of modifications to ensure contiguity of numpy arrays in some special cases where an evaluation (and hence copy, to contiguous memory) did not occur (when not using numexpr and when the expression was simply "o0"). Consequently also had to improve the rebasing of operands which the lazyexpr machinery carries out (prior to this PR, the rebasing would not occur when the lazyexpr was just a blosc2 array, i.e. "a" would not be rebased to "o0").
Before merging have to do a benchmark comparing new slicing with non-unit steps for lazy expressions (which only uses numpy at chunk-level) with the old format (which filled the whole span of the slice before taking the steps at the very end). Consequently, the old version might load a very large result into memory, even if the slice was small - consider an array of shape ``(10**9, )`` with slice ``[::10**8]``; the old Blosc2 would fill an array of shape ``(9.9**9, )`` and then slice to give a result of shape (10,). The new version would never have more than the chunksize in memory.

### DIFF
--- a/tests/ndarray/test_lazyexpr.py
+++ b/tests/ndarray/test_lazyexpr.py
@@ -1089,9 +1089,7 @@ def test_eval_getitem2():
     np.testing.assert_allclose(expr[1:, :7], nres[1:, :7])
     np.testing.assert_allclose(expr[0:10:2], nres[0:10:2])
     # Now relies on inefficient blosc2.ndarray.slice for non-unit steps but only per chunk (not for whole result)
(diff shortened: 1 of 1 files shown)


lshaw8317:
pre-commit.ci autofix

PR #449, opened 2025-08-05
Pythagorean calculation with sub-optimal numerics

DimitriPapadopoulos:


examples/ndarray/animated_plot.py
```diff
@@ -54,7 +54,7 @@ def get_memory_mb():
 # --- 3D Data Generation ---
 def compute_3Ddata():
     time_factor = X * Y * 0.001
-    R = np.sqrt(Y**2 + Z**2)
+    R = np.hypot(Y, Z)
```
FrancescAlted, line 57: That's a good point.  However, this does not work with blosc2 (yet) because it is not implemented in numexpr.  In addition, performance with numpy is basically the same.

tests/ndarray/test_lazyexpr.py
```diff
@@ -1509,7 +1509,7 @@ def test_complex_lazy_expression_multiplication():
     time_factor = 0.5
 
     # First complex expression: R * 4 - time_factor * 2
-    R = np.sqrt(X_b2**2 + Y_b2**2)
+    R = np.hypot(X_b2, Y_b2)
```
FrancescAlted, line 1512: Ditto

FrancescAlted (commented):
Suggested changes break things, so closing.

Issue #454, opened 2025-08-13
Add functionality to C2Array


lshaw8317:
Add attributes to acces e.g. cratio, nbytes etc. (to match as far as possible NDArray). 
Add slice method (follow logic of Caterva2 api_utils.fetch_data function) to return blosc2 arrays not just NumPy arrays.

PR #455, opened 2025-08-13
C2array

lshaw8317:
Partial solution of #454


FrancescAlted:
Is that ready for review, or should be marked as 'draft'?

Issue #441, opened 2025-07-25
Optimise fancy indexing further


lshaw8317:
Continue to fine tune functionality and performance of fancy indexing according to outline suggested in #437

lshaw8317:
For implementing fancy indexing in slice (i.e. return blosc array not numpy array), one should return compressed output of ``get_fselection`` if the result can fit (uncompressed) in memory. If not, will have to write new code which loops over chunks of output, filling them in (possibly use chunk cache to speed things up a bit).

PR #459, opened 2025-08-21
Fancy index

lshaw8317:
Edited code to try and make fancy indexing faster (avoid searching full index every time via ndindex:as_subidx) and incorporate 1D fast path better. Could maybe be revised to allow for interleaved slice and integer arrays (currently it follows ndindex in only allowing slices before and after the integer arrays). Improvements are not huge, but there is a performance benefit, compared to the old algorithm, which found intersecting chunks via ndindex, and looped over those chunks, converting the full index to a local chunk index at every iteration. This involved searching the whole index for every loop (it remains as the default algorithm for boolean indexing).

The algorithm adaps he Zarr vindexing implementation which uses a sorting-based algorithm. Intersecting chunks are found (for slices and integer arrays), the integer array is sorted (by chunks for N>1 integer arrays, by index for a single integer array), and then the code loops over the chunks, taking the indexes corresponding to the chunk from the sorted integer array. For a single integer array,  only the relevant part of the chunk (between the minimum and maximum index within a chunk) is loaded - for several integer arrays, because of how ``get_slice_numpy`` is written, it is necessary to load the whole chunk and then index on it.
 Hence on avoids repeated searches over the whole index. There is a small detail in using np.unique, which does not quite function like np.bincount, since it sorts and may do a copy, which can cause a slowdown if not handled correctly.

Improvements could be made to try and find a way to use np.bincount directly, as I still do not completely trust np.unique to be fast. Also one could rewrite to interleave slices and indices (matching numpy). ``ndindex.expand`` also seems to consume quite a bit of memory (via copying) when broadcasting indices, I think ``np.broadcast`` would just do a view so might be better. Finally, one could rewrite ``get_slice_numpy`` to allow continuous, flat returns, and not necessarily only slices (a sequence of slices, which is multidimensional ''rectangular'', differs from a range from one multidimensional index to another, which is essentially 1D). See #441.

From the plot one can see that the new algorithm is just as fast as the old optimised 1D path, although it handles more cases.
newfancybench


bench/ndarray/fancy_index.py
```diff
@@ -37,9 +37,10 @@ def genarray(r, ndims=2, verbose=True):
     shape = (d,) * ndims
     chunks = (d // 4,) * ndims
     blocks = (max(d // 10, 1),) * ndims
+    urlpath = f'linspace{r}{ndims}D.b2nd'
     t = time.time()
     arr = blosc2.linspace(0, 1000, num=np.prod(shape), shape=shape, dtype=np.float64,
-                          urlpath=f'linspace{r}{ndims}D.b2nd', mode='w')
+                        urlpath=urlpath, mode='w')
```
FrancescAlted, line 43: Bad alignment?

bench/ndarray/fancy_index.py
```diff
@@ -136,16 +137,17 @@ def timer(arr):
         error_kw=dict(lw=2, capthick=2, ecolor='k'))
         labs+=label
 
-filename = f"results{labs}{NDIMS}D" + "sparse" if SPARSE else f"results{labs}{NDIMS}D"
+filename = f"{labs}{NDIMS}D" + "sparse" if SPARSE else f"{labs}{NDIMS}D"
+filename+=blosc2.__version__.replace('.','_')
```
FrancescAlted, line 141: PEP8: always surround operands (`+=`) between spaces. Are you using ruff pre-commit?

src/blosc2/ndarray.py
```diff
@@ -4519,3 +4572,79 @@ def __setitem__(self, selection, input) -> np.ndarray:
 
 #     def __setitem__(self, selection, input) -> np.ndarray:
 #         return NotImplementedError
+
+
+def slice_to_chunktuple(s, n):
+    """
+        # credit to ndindex for this function #
```
FrancescAlted, line 4579: Missed TODO action?

tests/ndarray/test_ndarray.py
```diff
@@ -292,34 +292,43 @@ def test_oindex():
     np.testing.assert_allclose(arr[:], nparr)
 
 
-def test_findex():
+@pytest.mark.parametrize("c", [None, 10])
+def test_findex(c):
```
FrancescAlted, line 296: Try to use slightly more verbose names, like `test_fancy_index`.

tests/ndarray/test_ndarray.py
```diff
@@ -292,34 +292,43 @@ def test_oindex():
...
     rng = np.random.default_rng()
     idx = rng.integers(low=0, high=d, size=(100,))
 
     row = idx
     col = rng.permutation(idx)
     mask = rng.integers(low=0, high=2, size=(d,)) == 1
-
-    ## Test fancy indexing for different use cases
+    #
+    # ## Test fancy indexing for different use cases
```
FrancescAlted, line 328: Too many hashes?

tests/ndarray/test_ndarray.py
```diff
@@ -292,34 +292,43 @@ def test_oindex():
...
     m, M = np.min(idx), np.max(idx)
     nparr = arr[:]
-    # i)
+    # # i)
```
FrancescAlted, line 331: ditto

FrancescAlted (approved):
Look great to me.  It would be interesting to see how the new performance compares with h5py and zarr, for a reference.

PR #460, opened 2025-08-24
Add `as_ffi_ptr` to NDArray

barakugav:


FrancescAlted:
Ok. Can you explain why exactly you want to add this functionality?

src/blosc2/__init__.py
```diff
@@ -328,6 +328,7 @@ class Tuner(Enum):
     arctan,
     arctan2,
     arctanh,
+    array_from_ffi_ptr,
```
FrancescAlted, line 331: Better move this to the https://github.com/Blosc/python-blosc2/blob/main/src/blosc2/__init__.py#L106 block.

FrancescAlted: No, I think the current is the proper place, so no worries.

PR #467, opened 2025-09-12
Array-api 2

lshaw8317:
Benchmark matmul to check it's not slower than Ricardo's original implementation.
Also benchmark matmul vs tensordot (since tensordot is a generalised matmul).

### DIFF
--- a/tests/ndarray/test_getitem.py
+++ b/tests/ndarray/test_getitem.py
@@ -206,7 +206,7 @@ def test_take_along_axis(shape, chunkshape, axis):
     # Make some indices with same shape except for the given axis
     indices_shape = list(shape)
     indices_shape[axis] = 2  # we'll take 2 indices along that axis
-    rng = np.random.Generator()
+    rng = np.random.default_rng()
     indices = rng.integers(0, shape[axis], size=indices_shape)
 
     # NumPy expected


src/blosc2/ndarray.py
```diff
@@ -4575,60 +4585,189 @@ def matmul(x1: NDArray, x2: NDArray, **kwargs: Any) -> NDArray:
...
+
+    Parameters
+    ----------
+    x1: NDArray
+        First input array. Should have a numeric data type.
+
+    x2: NDArray
+        Second input array. Should have a numeric data type. Corresponding contracted axes of x1 and x2 must be equal.
+
+    axes: (Union[int, Tuple[Sequence[int], Sequence[int]]])
```
FrancescAlted, line 4668: Try to be consistent with the type declaration above. Also, I prefer syntax like `int | tuple[Sequence[int], Sequence[int]]` rather than more verbose `(Union[int, Tuple[Sequence[int], Sequence[int]]])`

src/blosc2/ndarray.py
```diff
@@ -385,7 +385,7 @@ def _check_allowed_dtypes(
     ):
         raise RuntimeError(
             "Expected LazyExpr, NDArray, NDField, C2Array, Proxy, np.ndarray or scalar instances"
-            f" and you provided a '{type(value)}' instance"
+            + f" and you provided a '{type(value)}' instance"
```
FrancescAlted, line 388: You don't need to add the `+`, as this is implicit when two strings are in different lines.

PR #469, opened 2025-09-14
Undefined name in `__all__`

DimitriPapadopoulos:


src/blosc2/__init__.py
```diff
@@ -521,7 +521,6 @@ def _raise(exc):
     "are_partitions_aligned",
     "are_partitions_behaved",
     "asarray",
-    "astypeclib_info",
```
FrancescAlted, line 524: This is a typo.  Fixed in 9f602d0

FrancescAlted (commented):
This has been fixed in 9f602d0

PR #479, opened 2025-09-18
Fix view

lshaw8317:
Prior to this fix c = blosc2.expand_dims(c, axis=1) would cause the reference to the original array c to disappear, meaning that the original data is garbage collected and the data to which the expanded view points no longer existed. Consequently, c[0] for example would fail.

By including a base attribute in NDArray objects we ensure views always hold a reference to the array object. When views are garbage collected (created) the reference count for the original array then automatically reduces (increases) by one.


FrancescAlted (approved):
Excellent work!

PR #482, opened 2025-09-19
Mark fancy indexing tests as heavy

FrancescAlted:
Fancy indexing tests take long to time to execute, making the current whole suite more than 30% slower; this is a bit too much for a test.  This marks the test as heavy, so that:

```
pytest tests/ndarray/test_ndarray.py -v -m "heavy"
```

can still execute them.


lshaw8317 (commented):
LGTM

PR #487, opened 2025-09-29
Array protocol

FrancescAlted:
This add a new Array protocol that is useful to describe the minimal interface that should be supported for blosc2 functions to deal with array inputs.  This protocol is really minimal, and it does not actually depend on any non-standard array API method or property (like `.chunks`); @lshaw8317 hope this is fine for most (if not all) the signature changes proposed here, but if not, we need to discuss which cases require a more complete interface (blosc2.ChunkedArray ?).

So instead of describing the input types as e.g. `NDArray | NDField | C2Array | LazyArray`, they are described now as `blosc2.Array`.

As a plus, and thanks to the `@runtime_checkable` decorator, compliance with the new Array protocol can be tested at runtime with `isinstance(object, blosc2.Array)`.


src/blosc2/lazyexpr.py
```diff
@@ -337,7 +337,7 @@ def compute(self, item: slice | list[slice] | None = None, **kwargs: Any) -> blo
         pass
 
     @abstractmethod
-    def __getitem__(self, item: int | slice | Sequence[slice]) -> blosc2.NDArray:
+    def __getitem__(self, item: int | slice | Sequence[slice]) -> blosc2.Array:
```
lshaw8317, line 340: Since this is getitem, shouldn't it return a NumPy Array?

src/blosc2/lazyexpr.py
```diff
@@ -3252,7 +3250,7 @@ def info(self):
     def info_items(self):
         inputs = {}
         for key, value in self.inputs_dict.items():
-            if isinstance(value, np.ndarray | blosc2.NDArray | blosc2.C2Array):
+            if isinstance(value, blosc2.Array | blosc2.C2Array):
```
lshaw8317, line 3253: C2Array fulfills the blosc2.Array interface, no?

src/blosc2/ndarray.py
```diff
@@ -367,43 +403,32 @@ def reshape(
...
-    value: bool
-    | int
-    | float
-    | str
-    | blosc2.NDArray
-    | blosc2.NDField
-    | blosc2.C2Array
-    | blosc2.Proxy
-    | blosc2.LazyExpr,
+    value: bool | int | float | str | blosc2.Array | blosc2.Proxy | blosc2.LazyExpr,
```
lshaw8317, line 406: LazyExpr is not blosc2.Array compliant?

src/blosc2/lazyexpr.py
FrancescAlted: Fixed in 28e560d

PR #488, opened 2025-09-29
Separate heavy tests in element wise testing

FrancescAlted:
The new `test_elementwise_funcs.py` module uses around 40s on my (relatively modern) laptop.  This PR allows to mark some tests as 'heavy'. After that:

```bash
> pytest tests/ndarray/test_elementwise_funcs.py
<snip>
============================================== 878 passed, 10 skipped, 444 deselected, 10 warnings in 9.71s ==============================================
```

```bash
> pytest tests/ndarray/test_elementwise_funcs.py -m heavy
============================================== 438 passed, 6 skipped, 888 deselected, 6 warnings in 30.04s ===============================================
```

which is more manageable, and compliant with the existing convention of marking heavy tests as 'heavy'.


lshaw8317 (commented):
LGTM

PR #511, opened 2025-10-21
Network tests have been migrated to new cat2.cloud/testing site

FrancescAlted:


tests/conftest.py
```diff
@@ -20,7 +20,7 @@ def pytest_configure(config):
 def c2sub_context():
```
lshaw8317: Maybe change this name now that we don't use pub/sub terminology but it's not really important

lshaw8317:
LGTM

PR #519, opened 2025-11-01
Fix Pylint issue

DimitriPapadopoulos:
Assigning result of a function call, where the function has no return


FrancescAlted:
LGTM. Thanks @DimitriPapadopoulos !

PR #524, opened 2025-11-11
Fix #522

### LINKED ISSUE #522: Branches of the `if` statement have identical implementation
**Describe the bug**

This piece of code introduced by 48724b9 / #423 doesn't make sense:

https://github.com/Blosc/python-blosc2/blob/3816b864ad4c492733b11bcde0c405752f6ba62f/bench/ndarray/concatenate.py#L268-L272

The two branches share the same implementation.

lshaw8317:
Correct issue raised in #522


FrancescAlted:
LGTM

PR #536, opened 2025-11-28
Set the number of threads to 2 to reduce the test suite runtime. Fixes #535.

### LINKED ISSUE #535: Using pytest >= 8 more than doubles test execution time
I noticed that, when updating to pytest >= 8, the test suite more than doubled time.

For pytest >= 8:

```
8193 passed, 7465 deselected in 242.41s (0:04:02)
```

For pytest < 8:

```
8193 passed, 7465 deselected in 95.24s (0:01:35)
```

After some digging, I noticed that pytest >= 8 executes the tests in `tests/ndarray` before the ones in `tests/`, so the conclusion is that some test in `tests/` made the others run faster.  After some attempts, I detected that `tests/test_storage.py::test_dparams_defaults` was setting the number of threads to 1 (`blosc2.set_nthreads(1)`) and that setting remained for the rest of test suite, making it faster.

This makes sense, as launching a large number of threads (typically the number of logical cores in local CPU) is a quite expensive thing to do for small tests like the ones in our suite.  It would be good to use a smaller number of threads to remove the big overhead that this is introducing (for CPUs with a large number of cores).

FrancescAlted:
With this patch, the runtime when using latest pytest (9.x), the time is quite contained:

```
8193 passed, 7465 deselected in 101.11s (0:01:41)
```

which is pretty close the time for using 1 thread for all tests in `tests/ndarray ` (95 sec in my box).

Using 2 threads combines testing the multithreading with contained run times for the test suite.

Closes #535

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,7 @@
 # LICENSE file in the root directory of this source tree)
 #######################################################################
 import os
+import sys
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


lshaw8317 (approved):
LGTM

PR #538, opened 2025-12-03
Allow saving of numba-decorated lazyudfs

lshaw8317:
Fix security issue with unrestricted builtins too.


FrancescAlted (approved):
Just a cosmetic suggestion.  Otherwise, LGTM.

PR #557, opened 2026-01-17
Update .gitignore file.

Nirkan:
Adds **blosc2** and **lazyarray.b2nd** to the **.gitignore** file. The files were generated locally during editable installation and testing.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,9 +64,11 @@ You can find examples at:
 
 https://github.com/Blosc/python-blosc2/tree/main/examples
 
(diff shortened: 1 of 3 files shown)


FrancescAlted:
Ok. I fixed the `lazyarray.b2nd` leftover file in eae604d4.  However, adding `blosc2` to .gitignore makes python wheels to fail (not sure why).

Nirkan:
Thank you for looking into it. I am not sure as well.

Issue #466, opened 2025-09-10
Array API compliance


lshaw8317:
The following is a summary of the failing tests in the array-api-tests suite. When you have managed to make the test pass, mark the box next to the file.

There are three main groups of tests:
1 - lazy expression and chunking related things (unique_counts, cumulative_sum etc.)
2 - indexing related things (chiefly implement fancy indexing for .slice method)
3 - improving numexpr (add op_codes such as ge_bcd etc.)

- [ ] ``test_utility_functions.py`` : add ``blosc2.diff`` function, ~``squeeze`` needs to be modified to return a view (at C-level, follow expand_dims)~
- [x] ~``test_statistical_functions.py`` : add ``cumulative_sum`` and ``cumulative_prod``~
- [ ] ``test_special_cases.py``: everything
- [ ] ``test_sorting_functions.py``: add ``argsort``
- [ ] ``test_signatures.py`` : everything
- [ ] ``test_set_functions.py``: add ``unique_counts``, ``unique_inverse``, ``unique_all``, ``unique_values``
- [ ] ``test_searching_functions.py``: ~add ``argmax``, ``argmin``~, ``where`` attribute to ``NDArray``, ``nonzero``, ``count_nonzero``
- [ ] ``test_operators_and_elementwise_functions.py``: everything
- [ ] ``test_manipulation_functions.py``: SERIOUS (Fatal Python error: Floating-point exception)
- [ ] ``test_linalg.py``: haven't implemented most things (see #476), which is fine, but some tests also fail because numexpr needs a ge_bcd opcode.
- [x] ~``test_inspection_functions``: passes all~
- [ ] ``test_indexing_functions.py``: need to improve``take`` and ``take_along_axis`` (latter fails tests)
- [ ] ``test_has_names`` : fails 23 tests, need to add functions listed in comment below
- [ ] ``test_fft.py``: skips all tests
- [ ] ``test_data_type_functions.py``:  a few different things (11 tests), add ``broadcast_arrays``
- [ ] ``test_creation_functions.py``:  add ``reshape`` for N-D arrays.
- [x] ~``test_constants.py``: passes all~
- [ ] ``test_array_object.py``: one test takes too long for some reason sometimes

In addition, it would be nice to revamp the documentation to mirror the array API layout and classification of functions/classes/methods etc.

In order to run the tests, follow these steps:
1. Git clone array-api test suite https://github.com/data-apis/array-api-tests
[...]

PR #558, opened 2026-01-19
Adding cumulative sum and product functions.

Nirkan:
This PR addressed part of the issue Array API Compliance #466  i.e  adding cumulative_sum and cumulative_prod to pass the test : array_api_tests/test_statistical_functions.py

Two functions (cumulative_sum and cumulative_prod) are added to ndarray.py and passes array_api_tests/test_statistical_functions.py

### DIFF
--- a/src/blosc2/__init__.py
+++ b/src/blosc2/__init__.py
@@ -434,6 +434,7 @@ from .ndarray import (
     save,
     stack,
     cumulative_sum,
+    cumulative_prod,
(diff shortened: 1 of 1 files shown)


lshaw8317:
Thanks for your contribution, however it would be better to use Blosc2's internal lazyexpr machinery. Please look at #514 for a roadmap to follow.
Also don't forget to add some tests please.

PR #565, opened 2026-01-23
Miniexpr for Windows

FrancescAlted:
Minimal modifications for allowing miniexpr to run on windows too.  It currently has these limitations:

* It does not work when the output of an expression is of integral type
* It does not work in expressions with operands with mixed types (e.g. int32 + float32 -> float64)

There has been a considerable amount of effort to reduce the limitations on windows to just this two cases, so I am quite satisfied with what we have now.  All other cases, including SIMD accelerations for transcendental functions via [SLEEF](https://sleef.org/index.xhtml), are working really well.

Ready for review.

### DIFF
--- a/src/blosc2/lazyexpr.py
+++ b/src/blosc2/lazyexpr.py
@@ -1352,7 +1352,7 @@ def fast_eval(  # noqa: C901
         res_eval = blosc2.uninit(shape, dtype, chunks=chunks, blocks=blocks, cparams=cparams, **kwargs)
         try:
             res_eval._set_pref_expr(expression, operands, fp_accuracy=fp_accuracy)
-            print("expr->miniexpr:", expression, fp_accuracy)
+            # print("expr->miniexpr:", expression, fp_accuracy)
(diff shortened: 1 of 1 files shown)


src/blosc2/lazyexpr.py
```diff
@@ -1326,14 +1325,34 @@ def fast_eval(  # noqa: C901
...
+                    use_miniexpr = False
 
     if use_miniexpr:
         cparams = kwargs.pop("cparams", blosc2.CParams())
         # All values will be overwritten, so we can use an uninitialized array
         res_eval = blosc2.uninit(shape, dtype, chunks=chunks, blocks=blocks, cparams=cparams, **kwargs)
         try:
             res_eval._set_pref_expr(expression, operands, fp_accuracy=fp_accuracy)
-            print("expr->miniexpr:", expression, fp_accuracy)
+            # print("expr->miniexpr:", expression, fp_accuracy)
```
lshaw8317: Should we remove this now that we are doing a release?

PR #570, opened 2026-02-05
Add string functionalities for strings in LazyExprs

lshaw8317:
Add support for unicode dtypes and following functions:

- [x] ==
- [x] startswith
- [x] endswith
- [x] contains

With miniexpr enabled, ``fast_eval`` is roughly 4x faster than when not using miniexpr (see ``bench/stringops_bench.py``):

```
== took 0.037s for miniexpr
== took 0.112s for normal fast path
contains took 0.037s for miniexpr
contains took 0.323s for normal fast path
startswith took 0.036s for miniexpr
startswith took 0.177s for normal fast path
endswith took 0.035s for miniexpr
endswith took 0.172s for normal fast path
```

Last thing to do is handle ``md_compile_nd_ex``/``md_compile_nd_jit`` problem, possibly at level of ``miniexpr``.

### DIFF
--- a/src/blosc2/ndarray.py
+++ b/src/blosc2/ndarray.py
@@ -4969,60 +4969,6 @@ def where(
     return condition.where(x, y)
 
 
-# def isin(
-#     element: str | int | float | blosc2.Array,
-#     test_elements: str | int | float | blosc2.Array,
-#     assume_unique=False,
-#     invert=False,
-#     kind=None,
-#     **kwargs,
-# ) -> NDArray:
-#     """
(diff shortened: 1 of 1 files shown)


FrancescAlted (commented):
Ok. This looks fine to me.  Remember to add a benchmark to see whether miniexpr can accelerate operations with strings or not.

lshaw8317:
> Ok. This looks fine to me. Remember to add a benchmark to see whether miniexpr can accelerate operations with strings or not.

It seems like miniexpr doesn't support strings for some reason...

PR #615, opened 2026-04-09
Check RELEASEGIL in SChunk.__init__, append_data, set_slice, and to_cframe

function2-llx:
## Summary

`set_releasegil(True)` releases the GIL for `compress2()`, `decompress2()`, `insert_data()`, and `update_data()`, but has no effect on `SChunk.__init__`, `append_data()`, `set_slice()`, or `to_cframe()`. Since `pack_tensor` / `pack_array2` go through `SChunk.__init__` + `to_cframe`, the primary serialization path holds the GIL for its entire duration — preventing any thread-level parallelism.

This PR applies the same pattern already used by `insert_data` and `update_data`: split `blosc2_schunk_append_buffer` (monolithic compress+append) into `blosc2_compress_ctx` (inside `with nogil:` when `RELEASEGIL` is set) followed by `blosc2_schunk_append_chunk` (GIL held). For `to_cframe`, `blosc2_schunk_to_buffer` is declared `nogil` and wrapped with the `RELEASEGIL` check.

## Benchmark

8 Python threads, `blosc2.set_nthreads(1)`, 200x512x512 int16 array, zstd clevel=3:

|  | serial | threaded (8w) | speedup |
|--|--------|---------------|---------|
| `compress2` (baseline) | 1.18 s/job | 0.21 s/job | 5.7x |
| `pack_tensor` **before** | 1.33 s/job | 1.47 s/job | 0.9x |
| `pack_tensor` **after** | 1.34 s/job | 0.25 s/job | 5.4x |

### 1. Check RELEASEGIL in SChunk.\_\_init\_\_, append_data, set_slice, and to_cframe

Purely mechanical — applies the `insert_data`/`update_data` GIL release pattern to all four methods:

- **`SChunk.__init__` data loop**: compress each chunk via `blosc2_compress_ctx` with GIL release, then `blosc2_schunk_append_chunk`
- **`append_data()`**: same transformation
- **`set_slice()`**: wrap existing `blosc2_compress_ctx` call in `RELEASEGIL` check; replace `blosc2_schunk_append_buffer` in append loop with compress + append_chunk
- **`to_cframe()`**: declare `blosc2_schunk_to_buffer` as `nogil`, wrap call with `RELEASEGIL` check

### 2. Fix pre-existing issues in set_slice error handling

Separate commit for correctness fixes noticed while working on `set_slice`:

- Set `current_nchunk` before `compress_ctx` in the update-chunk path (needed by prefilters, was missing unlike all other compress sites)
- Add `PyBuffer_Release` on error paths that were leaking the buffer
- Add `comp_rc == 0` and `_check_comp_length` guards for consistency with `insert_data`/`update_data`

## Test plan

- [x] All existing SChunk tests pass (773 passed, 0 failed)
- [x] Correctness verified with `RELEASEGIL=True`: serial and threaded (8w x 32 jobs) `pack_tensor` roundtrips produce identical arrays
- [x] `append_data`, `to_cframe`, `set_slice` roundtrips verified with `RELEASEGIL=True`

## Notes

- The new code follows the same `realloc(chunk, comp_size)` pattern used by `insert_data`/`update_data` without checking for `realloc` NULL return. Unreachable in practice (shrinking realloc), but could be hardened across all sites in a follow-up.
- Similarly, `_check_comp_length` can theoretically raise and leak the chunk buffer — same as in `insert_data`/`update_data`. Also unreachable (compress always produces >=16-byte headers).
- The new code adds `free(chunk)` on the `comp_size < 0` error path, which is missing in the existing `insert_data`/`update_data`.

### DIFF
--- a/src/blosc2/blosc2_ext.pyx
+++ b/src/blosc2/blosc2_ext.pyx
@@ -377,7 +377,7 @@ cdef extern from "blosc2.h":
     blosc2_schunk *blosc2_schunk_open_offset(const char* urlpath, int64_t offset)
     blosc2_schunk* blosc2_schunk_open_offset_udio(const char* urlpath, int64_t offset, const blosc2_io *udio)
 
-    int64_t blosc2_schunk_to_buffer(blosc2_schunk* schunk, uint8_t** cframe, c_bool* needs_free)
+    int64_t blosc2_schunk_to_buffer(blosc2_schunk* schunk, uint8_t** cframe, c_bool* needs_free) nogil
     void blosc2_schunk_avoid_cframe_free(blosc2_schunk *schunk, c_bool avoid_cframe_free)
     int64_t blosc2_schunk_to_file(blosc2_schunk* schunk, const char* urlpath)
     int64_t blosc2_schunk_free(blosc2_schunk *schunk)
@@ -1476,6 +1476,10 @@ cdef class SChunk:
         cdef int64_t index
         cdef Py_buffer buf
         cdef uint8_t *buf_ptr
+        cdef int comp_size
+        cdef int32_t csize
+        cdef uint8_t* chunk
+        cdef int32_t len_chunk
         if data is not None and len(data) > 0:
             PyObject_GetBuffer(data, &buf, PyBUF_SIMPLE)
             buf_ptr = <uint8_t *> buf.buf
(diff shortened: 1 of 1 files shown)


FrancescAlted (approved):
LGTM

FrancescAlted:
Thanks @function2-llx !

PR #622, opened 2026-04-20
CTable improvements

FrancescAlted:
### Motivation

 This PR extends CTable with first-class support for derived data:
 - virtual computed columns for convenience and expressiveness
 - materialized computed columns for persistence and stored-column indexing
 - direct expression indexes for planner-level acceleration without schema pollution
 - dedicated idioms for views

 ### Highlights

 - computed columns are now part of the CTable user model
 - materialized computed columns give a simple path to indexing computed results
 - direct expression indexing brings CTable closer to NDArray indexing capabilities
 - now, col[:] returns numpy values; col.view[...] returns a writable Column view

 ### Implementation notes

 - computed columns are stored as metadata and backed by lazy expressions
 - materialized columns are regular stored columns with recorded provenance
 - direct expression indexes are table-owned and backed by hidden physical expression-value arrays
 - where(...) and sort_by(...) can reuse matching expression/stored indexes when available
 - .to_numpy() has been removed in favor of `[...]`, which is more consistent with existing code.


Jacc4224 (approved):
LGTM

PR #639, opened 2026-05-19
Remove useless else clause detected on the loop

DimitriPapadopoulos:
When a loop specifies no break statement, the else clause will always execute when the loop sequence is empty, thus making it useless.


FrancescAlted:
Not sure why this is not passing tests.

FrancescAlted:
I think it is better to close this now.

DimitriPapadopoulos:
Yes, I think I had already tried that in the past. While the suggestion makes sense for pure Python code, it doesn't make sense when testing the DSL.

PR #683, opened 2026-07-25
Fix packing of 0-dim arrays

### LINKED ISSUE #680: pack_array2/pack_tensor reject 0-D Numpy Arrays since 3.0
```python
import blosc2
import numpy as np

value = np.array(17)
print(value.shape, value.size, value.nbytes)

packed = blosc2.pack_array2(value)
restored = blosc2.unpack_array2(packed)

assert restored.shape == ()
assert restored == value
```
I get `TypeError: len() of unsized object` instead of successful execution as in blosc2 2.7.1. 

Caused by https://github.com/Blosc/python-blosc2/commit/e5f527593419583e278417199da2ce2ae2905678, numpy intentionally throws when calling `len` on 0-d array. I think fix is using the already defined `buf.len`/`len_data` instead of `len(data)`

Zentrik:
Fixes https://github.com/Blosc/python-blosc2/issues/680.

I did use claude for this, but the change seems basically what I would have written by hand.

### DIFF
--- a/src/blosc2/blosc2_ext.pyx
+++ b/src/blosc2/blosc2_ext.pyx
@@ -1639,11 +1639,11 @@ cdef class SChunk:
         cdef int32_t csize
         cdef uint8_t* chunk
         cdef int32_t len_chunk
-        if data is not None and len(data) > 0:
+        if data is not None:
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Thanks @Zentrik !

PR #697, opened 2026-08-11
Support for abi3 and forthcoming abi3t

FrancescAlted:


FrancescAlted:
This seems goo to go. The worse slowdown is `NDArray full slice (control)` (0.364 ms vs 0.540 ms for ~1.5x), but this cannot be reproduced locally, and timings are typically around just 5% in performance so this has been a clear outlier.  This makes sense, as most of bottlenecks are in compressing/decompressing, and in I/O (be memory, disk or the network).

At any rate, reducing the number of wheels seems a good thing, so merging.

PR #698, opened 2026-08-11
Fix asarray when chunk dims are greater than shape dims

Zentrik:
`asarray()` leaves arrays over 16 MB unreadable when the chunk shape is greater than the shape:

```python
a = np.arange(146 * 23802, dtype=np.float64).reshape(146, 23802)
b = blosc2.asarray(a, chunks=(147, 23802), blocks=(1, 23802))
b[:]  # RuntimeError: Error while getting the buffer
```

This seems to be because `asarray` has a separate path for large arrays which calls `schunk.update_data()` on its fastpath if `are_partitions_behaved`. I have changed `are_partitions_behaved` to no longer return true for this example and tried to simplify the existing logic which I found hard to follow. I've tried to preserve the old behaviour whilst also checking that `size % unit == 0` if `size < unit`.

The other uses of `are_partitions_behaved` seem to live in `lazyexpr.py` and are of the form `behaved and result.shape == out.chunks and result.dtype == out.dtype` so I think they're unaffected.

### DIFF
--- a/src/blosc2/ndarray.py
+++ b/src/blosc2/ndarray.py
@@ -308,8 +308,6 @@ def are_partitions_behaved(shape, chunks, blocks):
     bool
         True if the partitions are well-behaved, False otherwise.
     """
-
-    # Spanning the innermost dimension keeps whole container rows; outer dimensions just tile evenly
     def check_contiguity(container, part):
         if container and container[-1] != part[-1]:
             return False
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Well seen, and the fix looks good to me.  Merging.  Thanks @Zentrik !

PR #705, opened 2026-09-04
Unify lazy remote caching for Caterva2 and fsspec

FrancescAlted:
This PR gives Caterva2 `URLPath` objects and fsspec URLs a consistent lazy-open and caching interface. Remote arrays can now use an in-memory cache, a managed persistent cache directory, or an explicitly named persistent cache file, while fetching only the chunks or blocks required by each access.

It also makes persistent caches independently reopenable, improves remote traffic accounting and HTTP metadata discovery, documents the two remote access routes in detail, and reduces the runtime of several tests in the default suite.

## Main changes

- Extend `blosc2.open(URLPath(...), lazy=True)` to return a `Proxy` over a `C2Array`, matching the existing lazy fsspec behavior.
- Support the same cache controls for lazy Caterva2 and fsspec arrays:
  - no cache argument: process-local in-memory cache
  - `cache_dir=...`: persistent cache with a derived filename
  - `cache_path=...`: persistent cache at an exact filename
- Deprecate `cache_storage` in favor of `cache_dir`, while retaining it as a warning-producing compatibility alias.
- Validate that `cache_storage`, `cache_dir`, and `cache_path` are mutually exclusive.
- Detect stale persistent caches from the remote source stamp and automatically rebuild them.
- Add `Proxy.cache_status`, reporting `"created"`, `"reused"`, or `"invalidated/rebuilt"` for managed persistent caches.
- Record the source kind and location in persistent proxy metadata so fsspec and Caterva2 sources can be reconstructed when a cache file is opened directly.
- Keep Caterva2 authentication out of portable cache metadata; credentials are resolved again from the active Caterva2 context.
- Avoid a redundant Caterva2 metadata refresh when the newly constructed `C2Array` has already supplied current metadata and its source stamp.
- Optimize lazy HTTP opening to obtain the frame metadata and HTTP validator in one range request. ETag is preferred, with Last-Modified and size as a fallback.
- Include opening metadata responses in `C2Array` and proxy traffic counters, making request and byte totals cover the full remote interaction.
- Install fsspec's HTTP dependencies through the `fsspec` optional extra.

## Documentation and examples

- Rewrite the remote arrays guide around the distinction between fsspec byte-range access and Caterva2 dataset access.
- Document supported remote object types, cache lifetimes, exact cache paths, independent cache reopening, invalidation, authentication isolation, traffic measurement, block-level reads, and custom transports.
- Expand the `C2Array`, `URLPath`, and `FsspecNDSource` reference documentation.
- Add `examples/fsspec-cat2-access.py`, which compares persistent lazy access to the same dataset through Caterva2 and fsspec/HTTPS and reports cache status, traffic, timing, and cache size.
- Update existing installation, tutorial, and fsspec examples to use `cache_dir`.


FrancescAlted:
Superseeded by #708

PR #709, opened 2026-09-14
Add Windows ARM64 build

ndabas:
As the title says. A few things to note:
- I removed the "Add LLVM to PATH" step because it's already on the PATH in all current Windows runner images.
- I removed the "Install MSVC" step because that's already installed.
- I restructured the architecture detection to prepare for ARM64, to mirror the [miniexpr CMakeLists.txt](https://github.com/Blosc/miniexpr/blob/main/CMakeLists.txt).
- miniexpr TCC JIT is turned off because they don't yet support Windows ARM64.
- zlib-ng currently fails to build on Windows ARM64 with ClangCl -- this has been fixed in https://github.com/zlib-ng/zlib-ng/pull/2274 but there has been no release yet. So I've turned off NEON optimizations on Windows ARM64 for now, which can be re-enabled when the fix lands in the next release.
- We can also remove the "Install Ninja" step because that's already installed on all runners across all platforms. Let me know if you would like me to do that as well.


FrancescAlted:
Interesting.  Not using NEON optimizations is a good first step. Would that cover producing wheels and uploading to PyPI too?

ndabas:
Yes, I've added the changes to the cibuildwheels workflow so it will produce all of the wheels and upload to PyPI.

The NEON optimizations actually work already if we use the develop branch of zlib-ng. The fix just hasn't landed in a release yet.

I see a build error in the CI run you kicked off which wasn't happening in the runs on my branch -- I'll investigate that.

ndabas:
Okay so the "Build and test on windows-latest (python 3.12)" job failed because of a transient error in installing `sccache` from Chocolatey -- I haven't modified that workflow (`build.yml`) in this PR. It should work if you retry.

We can potentially use https://github.com/mozilla-actions/sccache-action to install sccache instead.

FrancescAlted:
Thanks @ndabas. Let's merge this now to experiment with the platform.  Shout when a new zlig-ng release would be available.

ndabas:
Great, thanks for merging @FrancescAlted, I'll post an update when I get a notification about a zlib-ng release.

